PR #18, opened 2026-04-27
update quick start commads

Neelpatel1604:


Neelpatel1604:
Taking this changes to another pull

PR #24, opened 2026-04-30
Feat/add setup

het0814:
- Add setup video to readme using git lfs
- update tests to include async client similar to new auth method


Xenogents (approved):

Neelpatel1604:
@Xenogents , Please wait for linting check to pass before merge if failed , can fix in same pull so does  have to create new one for linting!

Issue #37, opened 2026-05-03
[BOUNTY $100] 🐜 Best-in-Class Integration: CrewAI + Memanto Agentic Memory


mjfekri:
Issue Description:
The Challenge: CrewAI agents are powerful, but they often struggle with "long-term amnesia" across different sessions or complex workflows. We want to see how Memanto—our open-source agentic memory layer—can provide a permanent, searchable, and context-aware memory for a multi-agent Crew.

The Goal: Build a compelling, real-world example of a CrewAI "Crew" that utilizes Memanto as its primary memory layer to maintain state, remember user preferences, or store long-term task outcomes.

✅ Submission Criteria:
Your submission must include:

A Working Repository/Script: A clean Python implementation using crewai and memanto.

The "Memory Test" Use Case: A scenario where the agent must recall information from a previous run or a different agent (e.g., a "Research Agent" stores findings in Memanto, and a "Writer Agent" retrieves them 24 hours later).

Visual Proof: A 30–60 second terminal recording (using Loom or Asciinema) or a high-quality GIF showing the memory retrieval in action.

The "How-To" README: A short guide on how to swap standard CrewAI memory for Memanto.

⭐ Bonus Points (Increases chances of winning):
Integration with Cursor or n8n.

Using Memanto to handle "contradictory" memories (e.g., updating an old fact with a new one).

A thread on X demonstrating the project (tag @moorcheh-ai).

🚀 How to Claim:
Star this repository (We are pushing for 1k stars this month—your support helps the project!).

Comment below with /attempt to let the community know you're working on it.

Submit your Pull Request to the examples/ folder in this repo.

BountyHub Guide:
1. Set Up Your AccountSign In: Visit the BountyHub homepage and sign in using your GitHub account.
Connect Stripe: Navigate to your dashboard, click on Payouts, and connect a Stripe account to receive funds.
2. Solve the Issue on GitHub. Implement the Solution: Fork the repository, clone it locally, and implement the fix or feature. Create a Pull Request (PR): Submit your PR to the original GitHub repository.
3. Submit Your Claim Import the PR: Go back to the specific bounty on BountyHub and click Submit Claim. Provide the Link: Paste the URL of your GitHub pull request into the claim form.
4. Payout and Finalization: Review & Merge: The bounty creator (or repository owner) reviews your work. The bounty is typically paid out once the PR is merged. 
[...]

PR #41, opened 2026-05-03
feat: add CrewAI + Memanto multi-agent memory example

Suraj-kumar00:
### Description
- CrewAI example using Memanto as shared persistent memory layer
- Research Agent stores findings, Writer Agent recalls them across sessions
- Includes contradiction handling demo and Cursor integration (bonus)

This PR closes issue #37

### Files
All in `examples/crewai-memory/`:  tools, agents, tasks, crew, 4 run scripts, README

### Demo

[Please find the Demo video here](https://github.com/user-attachments/assets/fc28dad1-d752-4af1-90d3-d29228dd7b1d)

### Whole Architecture
Architecture_Daigram

### Bonus: Cursor Integration
After running the CrewAI pipeline, you can access the same memories from Cursor:

#### Connect Cursor to Memanto globally (all projects)
`memanto connect cursor --global`

#### Or for a specific project directory
`memanto connect cursor --project-dir /path/to/your/project`

Screenshot 2026-05-06 at 6 55 26 PM

### Test plan
- [x] `python run_full_pipeline.py` — both agents end-to-end or
- [x] `python run_research.py` - Step 1: Research Agent stores findings in Memanto
- [x] `python run_writer.py` - Step 2: Writer Agent retrieves those memories in a NEW session
- [x] `python run_research.py` then `python run_writer.py` — cross-session persistence
- [x] `python run_contradiction.py` — conflict detection

### Changes since review

Addressed after feedback from PR review:

- **Dual memory systems resolved**: All Crew constructors now explicitly set
  `memory=False`, preventing CrewAI from injecting its own LanceDB-backed
  "Search memory" / "Save to memory" tools alongside Memanto tools.
- **Architectural rationale documented**: New README section explains why
  tool-based integration is the correct pattern for Memanto (CrewAI's
  `StorageBackend.search()` expects vector embeddings, but Memanto's API
  performs semantic search from natural language text — no clean bridge exists).
- **Namespace design documented**: Explains the shared agent ID pattern and
  how Memanto's scope system (`memanto_agent_{agent_id}`) provides isolation
  between different crews/projects.


Suraj-kumar00:
Hello @mjfekri,

Could you please review this?

Suraj-kumar00:
Hello @mjfekri, i know i haven't provided the video please allow me time I'll provide by the evening.

Suraj-kumar00:
Here is the X post: https://x.com/surajk_umar01/status/2052023332388921758

FYI: @mjfekri

Suraj-kumar00:
Hello @Xenogents, could you please review my PR?

Xenogents:
Hi Suraj! Sorry many pr's to go through, good job on reading through all the submission requirements! The integration is solid and I don't necessarily have suggestions because we don't have a clear idea of what a proper Memanto integration would look like I would like to point out some considerations.

**Multiple memory systems**: Because your solution provides Memanto memory operations as tools, it doesn't override crewai's existing LanceDB memory system causing this solution to effectively perform remember and recall operations to 2 different memory systems. This isn't necessarily a problem but special care has to be made regarding receiving duplicate information and to consider crewai use cases and where the memories are stored (crewai stores locally, memanto stores on cloud).

**Token costs**: An advantage of using a tool calling approach is that you can have the LLM determine memory types, confidence, and other metadata which would otherwise not be possible unless an LLM was invoked. It's a great way to utilize Memanto's 13 memory types. There are other ways to more deeply integrate with crewai such as overriding their native memory system and removing LanceDB entirely, this approach saves token costs, doesn't have to use tool calling and doesn't run into the problems with multiple memory systems. The downside of this approach is that it can't attach any additional metadata to the Memanto memories.

Because there are advantages and disadvantages to both sides its unclear which is the better approach, but as of right now there are probably latent issues with using duplicate memory systems which you may have to resolve.

Other things to note is that crewai has a filesystem hierarchy for storing memories which might be difficult to translate into memanto. The goal isn't to necessarily translate it over but to consider the use cases of crewai and what might be the best way to handle memories in memanto namespaces and what's worth implementing.

Suraj-kumar00:
Thanks for the detailed review! @Xenogents. I've pushed changes addressing each point:

**Dual memory systems**: All Crews now explicitly set `memory=False`. When `memory=True`, CrewAI auto-injects its own "**Search memory**" and "**Save to memory**" tools into every agent (backed by `LanceDB`). Setting it to False ensures Memanto is the sole memory layer, no duplicate storage or retrieval conflicts.

**Tool-based vs native backend**: I investigated overriding CrewAI's **StorageBackend** protocol directly. The blocker is that `StorageBackend.search()` receives a pre-computed **vector embedding,** by that point the original query text is lost. Memanto's API does semantic search from natural language text (via Moorcheh's information-theoretic search), not raw vectors. There's no clean way to bridge this without redundant embedding or degraded search quality. The tool-based approach also lets the LLM select the right memory type, confidence, and tags, which maps directly to **Memanto's 13 semantic types**. I've documented both approaches and the trade-offs in the README.

**Namespace design**: Both agents share a single Memanto agent ID (`crewai-research-team`) mapping to one namespace (`memanto_agent_crewai-research-team`). This is intentional for cross-agent memory sharing. Different crews/projects use different agent IDs for isolation. This mirrors how CrewAI scopes memory to /crew/{crew_name} but through Memanto's namespace system instead.

cc: @het0814 @mjfekri

Suraj-kumar00:
Also @Xenogents, can you or someone from the team add a reviewer to this PR?

Thanks!

Xenogents (approved):

mjfekri:
Congratulations. The bounty is approved

Xenogents:
Congratulations on winning the bounty! Just some additional thoughts, I realize now that CrewAI's StorageBackend receives a precomputed vector embedding which caused a lot of problems for the other integrations because they would perform search using vector embeddings on a text namespace. Moorcheh itself supports namespaces with raw vector embeddings but not Memanto, we would either need to embed the text into vectors as the data is being sent to moorcheh or use some sort of dummy embedding. Both solutions are a bit janky so I felt tool calling is the best way to approach integrating Memanto as a result.

I haven't done a full end to end review and test of your code so there might be some final tweaks I put in, but overall I think your pr is the only one that both is a valid integration and works as intended.

PR #445, opened 2026-05-12
[BOUNTY $100] 🐜 LangGraph + Memanto: Cross-Session Memory Integration (Complete + Documented)

extremeWu:
Closed — superseded by PR #461.


nagiexplorer88:
Two runtime issues to fix before this can satisfy the demo bounty:

1. `app.py:189` compiles the graph with `MemorySaver()`, but both demo calls at `app.py:208` and `app.py:221` call `graph.invoke(...)` without a configurable `thread_id`/checkpoint config. With a checkpointer attached, LangGraph requires that configurable checkpoint key at invocation time, so the README path can fail before Session 1 completes. Either remove the checkpointer for this deterministic two-call demo, or invoke with explicit per-session configs such as `config={"configurable": {"thread_id": "session-1"}}` and `session-2`.

2. `memanto_adapter.py:82-85` calls `self._client.answer(agent_id=self.agent_id, query=query)`, but the current `SdkClient.answer` API takes `question=...`, not `query=...`. Any user trying the adapter `reflect()` helper will get an unexpected-keyword runtime error. Change that call to `question=query`.

These are both on the documented/example surface, so a reviewer following the submitted files can hit them without reaching the actual cross-session memory proof.

extremeWu:
Thanks for the review! Both issues are fixed now:

1. ✅ Removed `MemorySaver()` checkpointer — the deterministic demo no longer needs thread_id config
2. ✅ Fixed `memanto_adapter.reflect()` — now passes `question=` instead of `query=`

The updated files are live on the branch.

Issue #639, opened 2026-06-05
[BOUNTY $100] 🐜 The Great Agentic Memory Showdown: Memanto Benchmarking & Evaluation Challenge


mjfekri:
<html>
<body>
[...]

PR #660, opened 2026-06-06
feat: Memanto vs LangChain benchmark (closes #639)

KHHH2312:
Adds benchmarking suite. Demonstrates Memanto's superior performance. Please see social post here: [TODO]

## Summary by CodeRabbit

* **Documentation**
  * Added a runnable benchmark example that compares two memory implementations. It includes setup (optional API keys), real vs simulated modes, how to run multiple iterations, per-run latency and accuracy metrics, aggregated statistics (mean/std/median), and a summary comparison of results.


Xenogents:
Thanks for your contributions! We're closing this PR as we aren't looking to compare to LangChain as it isn't a dedicated memory layer and not what we are trying to compare against.

Issue #397, opened 2026-05-11
[BOUNTY $100] 🐜 The Memanto + LangGraph Integration Challenge: Give Your Graph a Permanent Brain


mjfekri:
**Description:**
LangGraph is the gold standard for stateful agents, but managing long-term memory across disjointed sessions or massive graphs is still a hurdle. We want to see the best example of Memanto acting as the long-term memory layer for a LangGraph agent.

The Bounty: $100 USD (Paid via BountyHub)
The Winning Metric: The Merged PR with the highest Social Traction Score by June 1st 2026.

**🛠️ How to Enter:**
Star the Repo: Help us reach our 1,000-star goal!

Build: Create a LangGraph workflow (e.g., a customer support agent or a research assistant) that uses memanto to store and retrieve "memories" outside of the standard LangGraph state.

Submit PR: Fork this repo and add your code to /examples/langgraph-memanto.

Amplify: Post a video/thread of your agent on X, LinkedIn, or Reddit. Tag #Memanto and @moorcheh-ai. Include the link to your post in the PR description.

**📊 The "Social Traction" Formula (Automated Check):**
To keep this objective and low-bandwidth for our team, we will only audit the top 5 PRs based on these metrics:

X (Twitter): Likes (1pt) + RTs (3pts) + Bookmarks (3pts).

GitHub: 👍 or 🚀 reactions on your PR (2pts each).

Reddit post (5pts) and each upvote +(2pts)

We use a simple script to verify these at the end of the deadline. No manual tracking required from you—just focus on building and sharing the best and most practical examples!

**✅ Technical Criteria:**
Must demonstrate Cross-Session Recall (The agent remembers something from "yesterday" that isn't in the current thread's state).

Clean, documented code in a single folder.

A 30-second GIF or video link in the README.md.

laolaoqi:
Hi! I'd like to take on this LangGraph + Memanto integration challenge.

**What I can do:**
- Build a production-grade LangGraph agent (customer support or research assistant) with Memanto as the persistent memory layer
- Implement cross-session recall - agent remembers context from prior sessions outside standard LangGraph state
- Clean, well-documented code with a 30-second demo GIF in README
- Post a demo video/thread on X/Twitter for social traction

**Timeline:** PR ready within 7 days of assignment. Social amplification will follow immediately post-PR.

[...]

PR #685, opened 2026-06-07
fix: revised solution for #397

JirA44:
Revised implementation addressing the review feedback.

not a developer with a GitHub account
- Use proper nouns like Memanto and LangGraph as needed
- Avoid any technical jargon
- Don't mention the original code files or any specific implementation details
- Keep it short and to the point

Here's the pull request description:
The issue was that the existing code didn't include the necessary integration logic for Memanto and LangGraph.  To fix this, we need to create a new graph interpreter that can handle both memory and knowledge data.  This involves defining a new node type for memory and a new edge type for knowledge, then implementing a parser that reads both types of data and builds a new graph structure.  The code will include a class for memory nodes, a class for knowledge edges, and a parser that reads both types of data and creates a new graph.  This implementation will allow the integration of Memanto and LangGraph into the existing graph system without modifying the original code.

Closes #397

## Summary by CodeRabbit

* **New Features**
  * Introduced a new example module demonstrating LangGraph integration with Memanto, providing a foundation for building graph-based applications with message handling capabilities.


JirA44:
Closing to submit a corrected version — the current implementation has quality issues that need to be addressed first.

PR #699, opened 2026-06-08
fix(upload): sanitize filename to prevent path traversal (CWE-22)

andesyteoss:
## Vulnerability Summary

The `upload_file` endpoint in `memanto/app/routes/memory.py` is vulnerable to **path traversal (CWE-22)** via the multipart upload filename.

Starlette's `MultiPartParser` passes the raw `Content-Disposition` filename header directly through to `UploadFile.filename` without any sanitization. Before this fix, the endpoint used that value directly in `os.path.join(tmp_dir, original_name)` to write the uploaded file contents to disk. An attacker who supplies a filename like `../../../tmp/evil.txt` can write attacker-controlled content outside the intended temporary directory.

The escaped file also persists after the `shutil.rmtree(tmp_dir)` cleanup runs, since it was written outside `tmp_dir`.

**Affected function:** `upload_file()` at line ~302 of `memanto/app/routes/memory.py`
**Route:** `POST /api/v2/agents/{agent_id}/upload-file`
**Severity:** Moderate (CWE-22 — Improper Limitation of a Pathname to a Restricted Directory)

### Preconditions

- Network access to the Memanto API server
- A valid session token (obtainable by calling `POST /api/v2/agents/{agent_id}/activate`, which requires only the server-wide `MOORCHEH_API_KEY` — no per-client authentication)
- Cloud backend configured (`on-prem` backend blocks this endpoint with a 501)

## Proof of Concept

After activating an agent session, an attacker sends a multipart upload where the filename in the `Content-Disposition` header contains path traversal sequences:

```bash
# 1. Create an agent and get a session token
curl -s -X POST http://localhost:8765/api/v2/agents \
  -H "Authorization: Bearer $MOORCHEH_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"agent_id": "test-agent"}'

TOKEN=$(curl -s -X POST http://localhost:8765/api/v2/agents/test-agent/activate \
  -H "Authorization: Bearer $MOORCHEH_API_KEY" | python3 -c "import sys,json; print(json.load(sys.stdin)['session_token'])")

# 2. Upload a file with a traversal filename
# The ../../../ prefix escapes the tmpdir and writes to /tmp/pwned.txt
curl -X POST http://localhost:8765/api/v2/agents/test-agent/upload-file \
  -H "Authorization: Bearer $MOORCHEH_API_KEY" \
  -H "X-Session-Token: $TOKEN" \
  -F "file=@/dev/stdin;filename=../../../tmp/pwned.txt;type=text/plain" <<< "attacker controlled content"

# 3. Verify the file was written outside the temp directory
cat /tmp/pwned.txt
# Output: attacker controlled content
```

The file `/tmp/pwned.txt` is written outside the application's temporary directory and survives the `shutil.rmtree()` cleanup.

## Fix Description

The fix applies two layers of defense:

**1. Filename sanitization (primary fix, line 350):**
```python
original_name = Path(file.filename or "upload").name
```
`Path.name` strips all directory components, so `../../../tmp/evil.txt` becomes just `evil.txt`. An additional guard handles edge cases where the result is empty or a bare `.`/`..`.

**2. Realpath containment check (defense-in-depth, line 371):**
```python
if not os.path.realpath(tmp_path).startswith(os.path.realpath(tmp_dir)):
    raise HTTPException(status_code=400, detail="Invalid filename")
```
Even if a future code change weakens the first check, this second layer verifies the resolved path stays within the temporary directory before any file I/O occurs.

Both layers are minimal changes to the existing code and don't alter the endpoint's behavior for legitimate filenames.

## Testing

The PR includes `tests/test_cwe22_path_traversal.py` with 20 tests covering:

- **Unit tests for sanitization logic:** Verifies `Path.name` correctly strips traversal sequences (`../../../etc/passwd` → `passwd`), absolute paths (`/etc/passwd` → `passwd`), and edge cases (`None`, empty string, `.`, `..`, `/`).
- **Realpath guard tests:** Confirms the defense-in-depth check correctly rejects paths that resolve outside the temp directory while allowing safe paths.
- **Integration tests via ASGI client:** Sends actual multipart requests through the FastAPI app with traversal filenames and verifies the returned `file_name` is sanitized, and that normal uploads continue to work.

All existing tests remain unaffected — the fix only narrows behavior for malicious filenames.

## Adversarial Review

Before submitting, we verified this is genuinely exploitable and not mitigated by existing controls. The `upload_file` endpoint requires a session token via `X-Session-Token`, but session tokens are obtainable by any caller who knows the server's `MOORCHEH_API_KEY` — the `activate_agent` endpoint performs no per-client authentication. The extension allowlist (`.pdf`, `.docx`, `.txt`, etc.) constrains which suffixes can be used but does not prevent the path traversal itself — `../../malicious.txt` passes the `.txt` extension check while still escaping the directory. Docker deployments run as a non-root user, which limits writable targets, but the vulnerability still provides a file-write primitive that did not exist before, distinct from what session access alone grants.

---

## Bug Fixes
* Strengthened file upload security with enhanced filename validation to prevent directory traversal attacks and unsafe path inputs.


sebastiondev:
Thanks for catching this, @coderabbitai — the feedback is valid.

The `startswith` check without a trailing `os.sep` could theoretically allow a prefix-matching bypass (e.g. `/tmp/abc` matching `/tmp/abcdef/file`). While the practical risk is low given the prior `basename` sanitization and use of a freshly-created temp directory, a defense-in-depth check should be robust.

**Fixed in 8fcd7fa:**
- Appended `+ os.sep` to the `realpath` prefix check in `memanto/app/routes/memory.py` (line 372)
- Updated the corresponding unit tests in `tests/test_cwe22_path_traversal.py` to match

```python
# Before
if not os.path.realpath(tmp_path).startswith(os.path.realpath(tmp_dir)):

# After
if not os.path.realpath(tmp_path).startswith(os.path.realpath(tmp_dir) + os.sep):
```

Verified locally that:
1. Safe paths (e.g. `report.pdf` within `tmp_dir`) still pass ✅
2. Traversal paths (e.g. `../../etc/passwd`) are still rejected ✅
3. Prefix-matching bypass (e.g. `/tmp/abcdef/file` vs `/tmp/abc`) is now correctly rejected ✅

sebastiondev:
Closing this as inactive — no maintainer response after 14 days.

The security finding and fix remain valid. If this is still relevant, I'm happy to reopen, rebase, or re-submit against a different branch. Just drop a comment.

Xenogents (approved):

het0814 (approved):
LGTM!

PR #707, opened 2026-06-08
fix(ui): remove plaintext API key and session token from /api/ui/config response

andesyteoss:
## Vulnerability Summary

**CWE-200: Exposure of Sensitive Information to an Unauthorized Actor**

The `GET /api/ui/config` endpoint in `memanto/app/ui/routes/ui_router.py` returns the plaintext Moorcheh API key (`api_key`) and the active session JWT (`session_token`) in its JSON response body. This endpoint has no authentication — the `APIRouter()` at line 22 is created with no `dependencies`, and `main.py:92` mounts it with none either. Combined with the default bind address of `0.0.0.0` (config.py:112), CORS origins set to `["*"]` (config.py:117), Docker port mapping `8000:8000`, and no mandatory reverse proxy, any network-reachable client can extract the server's cloud API key with a single GET request.

The API key (prefix `mk_`) grants direct access to the Moorcheh cloud backend — memory storage, agent operations, etc. — independent of the Memanto server. This means an attacker who extracts the key can use it from anywhere, even after losing network access to the Memanto instance, until the key is manually rotated.

**Affected file:** `memanto/app/ui/routes/ui_router.py`, function `get_ui_config()` (line 32)

**Data flow:** `ConfigManager.get_api_key()` reads `MOORCHEH_API_KEY` from `~/.memanto/.env` → returned as `"api_key": api_key` in the JSON response → delivered to any HTTP client without authentication.

## Proof of Concept

Before the fix, a single curl command extracts the full API key and session token:

```bash
# From any machine that can reach the Memanto server
curl -s http://<memanto-host>:8000/api/ui/config | python3 -c "
import sys, json
data = json.load(sys.stdin)
print('API Key:', data.get('api_key', '(not present)'))
print('Session Token:', data.get('session_token', '(not present)'))
"
```

Pre-fix output:
```
API Key: mk_abc123...full_key_here
Session Token: eyJhbGciOiJI...full_jwt_here
```

Post-fix output:
```
API Key: (not present)
Session Token: (not present)
```

The attacker can then use the extracted key directly against the Moorcheh API:

```bash
curl -H "Authorization: Bearer mk_abc123...full_key_here" \
     https://api.moorcheh.ai/v2/agents
```

## Fix Description

This PR removes two fields from the `/api/ui/config` response:

1. **`"api_key"`** — the raw plaintext API key. The response already contains `"api_key_configured"` (boolean) and `"api_key_preview"` (last 6 characters masked), which are sufficient for the UI to display key status without exposing the secret.

2. **`"session_token"`** — the raw active session JWT. The response already contains `"has_active_session"` (boolean), which tells the UI whether a session is active without leaking the token.

The change is minimal (2 lines removed) and preserves all existing safe metadata fields. The UI's `api_key_configured`, `api_key_preview`, and `has_active_session` fields continue to work as before, providing the frontend with the information it needs to render status indicators.

## Testing

A regression test file `tests/test_cwe200_api_key_leak.py` is included with three test cases:

1. **`test_config_endpoint_does_not_return_raw_api_key`** — verifies the `api_key` field is absent from the response.
2. **`test_config_endpoint_still_has_api_key_status_fields`** — verifies `api_key_configured` and `api_key_preview` are still present and the preview does not contain the full key.
3. **`test_config_endpoint_does_not_return_session_token`** — verifies the `session_token` field is absent.

The tests mock `ConfigManager` so they don't require a real API key or disk access. They are designed to fail on the pre-fix code and pass on the post-fix code.

## Adversarial Review

Before submitting, we attempted to disprove this finding. We checked whether any framework-level authentication, middleware, or network restriction would prevent exploitation. The `APIRouter()` has no `dependencies` parameter, `main.py` adds no auth dependencies when including the router, the server binds to `0.0.0.0` by default, CORS allows all origins (`["*"]`), and the Docker configuration maps port 8000 directly. We also checked whether extracting the API key provides any capability beyond what network access to the Memanto server already grants — it does, because the `mk_` key works directly against the Moorcheh cloud API from any location, surviving loss of network access to the Memanto instance.

## Notes for Maintainers

- The broader unauthenticated API surface (e.g., `PUT /api/ui/api-key`, `GET /api/ui/browse`, `/api/v2/agents` lifecycle routes) is a separate architectural concern and not addressed here. This PR focuses narrowly on preventing plaintext secret leakage from the config endpoint.
- The UI static JavaScript references `S.config.api_key` and `S.config.session_token` — these will now be `undefined`. The UI already has the safe alternatives (`api_key_configured`, `api_key_preview`, `has_active_session`) available. If any UI flow depends on the raw values, it should be refactored to use server-side proxied endpoints instead.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the configuration endpoint to mask the API key value in the response, exposing only a truncated preview and related safe status fields while still returning session-related fields.

* **Tests**
  * Added contract tests to verify the API key is not returned in raw form and that the response includes the expected metadata (API key configured status, masked preview) and session token information.


sebastiondev:
Good catch — you're right that `test_config_endpoint_still_has_api_key_status_fields` was only validating the `api_key_configured` / `api_key_preview` safe fields but was missing the `has_active_session` assertion. Since the mock fixture returns `("agent-1", "tok_abc")` for `get_active_session()`, the endpoint should produce `has_active_session: True`, and we should be verifying that.

Added in 69b3b84:

```python
# Session status field should be present (replaces sensitive session_token)
assert "has_active_session" in data
assert data["has_active_session"] is True
```

All 3 tests pass.

Xenogents (approved):

sebastiondev:
Thanks for the merge. Glad it landed cleanly.

PR #743, opened 2026-06-15
fix: use same on prem moorcheh data if samenamespace already exist on…

het0814:
- update on-prem to use namespace is exist on agent creation ,instead of error 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved agent creation error handling by more reliably detecting “namespace already exists” conditions, so setup continues instead of failing unnecessarily.
  * Fixed memory search behavior for `min_similarity_score = 0.0` by preventing kiosk/threshold filtering from being applied when no positive similarity threshold is requested.

### DIFF
--- a/memanto/app/services/agent_service.py
+++ b/memanto/app/services/agent_service.py
@@ -79,10 +79,16 @@ class AgentService:
             # Namespace already exists - this is OK, agent might have been created before
             print(f"[OK] Namespace already exists in Moorcheh: {namespace}")
         except Exception as e:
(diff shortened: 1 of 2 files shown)


Neelpatel1604 (changes requested):
@het0814 Thanks for making changes, right now we are checking threshold is more than 0, can add threshold >0 AND threshold <= 1

threshold aka min_similariy

Neelpatel1604 (approved):

PR #776, opened 2026-06-24
fix(cli): honor custom title for short memories

cwjcwj95:
## Summary

Fixes a CLI title-resolution bug in `memanto remember`: when a user passed
`--title` with short memory content, the custom title was ignored and the memory
content was used as the title instead.

## Root Cause

The expression that resolves the title mixed `or` with a conditional expression:

```python
title or content[:50] + "..." if len(content) > 50 else content
```

Because the conditional expression binds around the whole `title or ...` branch,
short content always resolved to `content`, even when `title` was explicitly set.

## Fix

Wrap the default-title expression in parentheses so an explicit `--title` always
wins:

```python
title or (content[:50] + "..." if len(content) > 50 else content)
```

## Test Coverage

Added a CLI regression test covering short content with an explicit custom title:

```bash
pytest tests/test_cli.py -q -k "remember_short_content_honors_custom_title or test_remember"
```

Local environment note: this machine did not have `pytest` installed initially;
I verified syntax with `python3 -m py_compile memanto/cli/commands/memory.py tests/test_cli.py`
and `git diff --check`. A minimal dependency install was attempted but stopped
after it exceeded the time box.

## Bounty Context

This is a small reproducible usability/logic bug found while reviewing
`moorcheh-ai/memanto#770`. It is not a security disclosure and does not include
exploit details.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clarified the default title-selection logic for the `remember` command to ensure consistent evaluation order when falling back to a truncated content-based title.
* **Tests**
  * Added an integration test confirming the `remember` command honors an explicitly provided `--title` and triggers the expected client call.
  * Minor formatting adjustments to an existing unit test for improved readability.


het0814 (approved):

het0814:
Thanks for pointing out this small reproducible usability/logic bug. This is a good change though not a big enough issue to be considered for the bounty, thanks for your contributions to Memanto!

PR #777, opened 2026-06-24
fix: remove predictable session secret and repair security tests

leo1987820:
Related to #770

## Flaw summary

This PR now covers two reproducible security/test issues found while working through the Memanto Bug & Exploit Challenge.

### 1. Predictable default JWT session secret

`SessionService` falls back to the publicly known string `memanto-default-secret-change-in-production` when no secret is configured. The default settings layer also embedded that same value, so the shared `get_session_service()` path could inject a predictable signing secret even before `SessionService` fallback logic ran.

Impact: a deployment without an explicit `MEMANTO_SECRET_KEY` can sign sessions with a known secret, making forged JWT session tokens realistic.

Fix:
- Make `Settings.MEMANTO_SECRET_KEY` default to an empty string instead of a public secret.
- Generate a per-instance 256-bit fallback secret with `secrets.token_hex(32)` when no explicit secret is configured.
- Add regression tests proving the settings default no longer embeds the public value and missing secrets produce unique fallbacks.

### 2. CWE-200 API test setup regression

`TestCWE200ApiKeyLeak` referenced `self.TEST_AGENT_ID`, but that constant was only defined on a different test class. The two upload-file filename sanitization tests failed with `AttributeError` during setup instead of validating the intended path sanitization behavior.

Fix:
- Add a class-local `TEST_AGENT_ID` for `TestCWE200ApiKeyLeak` so the security tests exercise the intended assertions.

## Verification

- `.\.venv\Scripts\python.exe -m pytest tests/test_unit.py::TestSessionService::test_settings_default_does_not_embed_public_jwt_secret tests/test_unit.py::TestSessionService::test_missing_session_secret_generates_unique_fallback -q` -> failed before the fix, then passed after the fix.
- `.\.venv\Scripts\python.exe -m pytest tests/test_unit.py tests/test_api.py::TestCWE200ApiKeyLeak -q` -> `29 passed`.
- `.\.venv\Scripts\python.exe -m pytest -q` -> full local suite passed, with live Moorcheh E2E tests skipped because no real `MOORCHEH_API_KEY` is configured locally.
- `git diff --check` -> no whitespace errors.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved secret key handling: removed the known default; `MEMANTO_SECRET_KEY` now defaults to empty when unset, and session signing uses a securely generated per-instance fallback.
  * Strengthened endpoint validation: UI and session APIs now reject traversal/newline/glob-style `agent_id` and traversal-style dates with **400** across daily-summary and conflict-related routes.
  * Output safety: daily-summary now ignores any client-supplied `output_path` (always forwarded as `None`).
* **Tests**
  * Added regression and unit coverage for the validation rules, `output_path` ignoring, and secret-key fallback behavior (including uniqueness).


leo1987820:
Thanks for the duplicate check. I reviewed the two referenced PRs and wanted to clarify the remaining overlap/difference for maintainers:

- #1304 overlaps on the predictable default JWT secret, but takes a broader multi-fix approach. This PR keeps the JWT-secret change narrower by removing the public settings default and generating a per-instance fallback secret when no explicit key is configured.
- #1296 appears to cover a different session issue: automatic renewal replacing the persisted session without returning the replacement token to clients. That is separate from this PR's default-secret and summary/output-path hardening changes.
- This PR also includes the follow-up hardening from the earlier CodeRabbit review: strict summary key validation and ignoring client-supplied `output_path` for daily summary generation, with regression coverage for those paths.

So #777 is not a full duplicate of #1296, and only partially overlaps #1304 on the JWT-secret finding. The current branch is still intended as a smaller, focused patch with tests for the secret fallback plus the summary/output-path security regressions.

het0814:
Hi @leo1987820 

PR #798, opened 2026-06-25
Fix multi-scope memory search import

Ahmed-eae:
Summary
- Fix the `search_multi_scope` import so it loads `ScopeType` from the package's actual constants module.
- Add a regression test that verifies multi-scope searches build the expected namespaces and call the similarity backend.

Tests
- `pytest tests/test_memory_read_service.py`

Related: #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a path issue in memory search so scoped searches resolve correctly again.
* **Tests**
  * Added coverage for multi-scope memory search, verifying the search request uses the expected namespaces and returns results as expected.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -175,7 +175,7 @@ class MemoryReadService:
             namespaces = []
             from typing import cast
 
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #800, opened 2026-06-25
Fix namespace parsing for underscored scope IDs

n4p68vc6p9-hub:
## Summary

Fix a namespace round-trip bug for valid scope IDs that contain underscores.

## Bug / Impact

Memanto generates namespaces as `memanto_{scope_type}_{scope_id}`. For an agent ID such as `claude_dev`, `MemoryScope.to_namespace()` produces `memanto_agent_claude_dev`.

Before this change, `MemoryScope.from_namespace()` used `namespace.split("_")` and required exactly three parts, so the namespace generated by Memanto itself raised:

`ValueError: Invalid MEMANTO namespace format: memanto_agent_claude_dev`

This can break memory scope parsing for realistic agent, workspace, or session identifiers that use underscores, affecting memory isolation and namespace introspection.

## Reproduction

```bash
python -c "from memanto.app.core import MemoryScope; s=MemoryScope(scope_type='agent', scope_id='claude_dev'); print(s.to_namespace()); print(MemoryScope.from_namespace(s.to_namespace()))"
```

## Fix

Parse only the first two separators with `split("_", 2)`, preserving the full `scope_id` segment.

## Tests

- `python -m pytest tests\test_unit.py::TestMEMANTOArchitecture -q`
- `python -m pytest tests\test_unit.py -q`

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved namespace handling so scope IDs with underscores are preserved correctly when converting to and from the app’s namespace format.
  * Added coverage to confirm namespaces round-trip accurately for IDs containing underscores.

### DIFF
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -34,7 +34,7 @@ class MemoryScope(BaseModel):
         """Parse namespace back to scope"""
         from typing import cast
 
-        parts = namespace.split("_")
+        parts = namespace.split("_", 2)
         if len(parts) != 3 or parts[0] != "memanto":
             raise ValueError(f"Invalid MEMANTO namespace format: {namespace}")
         return cls(scope_type=cast(ScopeType, parts[1]), scope_id=parts[2])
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #803, opened 2026-06-25
Fix multi-scope search import path

n4p68vc6p9-hub:
## Summary
- fix multi-scope search to import `ScopeType` from the package path used elsewhere
- enforce the same TTL expiration filtering for multi-scope search results as single-scope search
- add regression coverage for namespace building and expired-result filtering

## Tests
- `python -m pytest tests/test_unit.py::TestMemoryReadService -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the configuration source used for multi-scope memory searches.
  * Improved multi-scope search to properly exclude expired memories from returned results and respect the requested limit.
* **Tests**
  * Added unit tests covering multi-scope namespace construction and ensuring the similarity query is executed once per search.
  * Added tests to verify TTL filtering, correct total counts, and that the search fetches enough candidates to account for post-filtering.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -175,7 +175,7 @@ class MemoryReadService:
             namespaces = []
             from typing import cast
 
-            from memanto.memanto.app.constants import ScopeType
+            from memanto.app.constants import ScopeType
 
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #808, opened 2026-06-25
Reject inactive session tokens

n4p68vc6p9-hub:
## Summary
- validate session JWTs against the persisted local session record
- reject tokens whose session was terminated, removed, or superseded by another session id
- add service and API regressions proving a deactivated token cannot write memories

## Tests
- `python -m pytest tests\test_unit.py::TestSessionService::test_validate_rejects_terminated_session_token -q`
- `python -m pytest tests\test_unit.py -q`
- `python -m pytest tests\test_api.py::TestMEMANTOAPI::test_deactivated_session_token_cannot_write_memory -q`
- `python -m pytest tests\test_unit.py tests\test_api.py::TestMEMANTOAPI -q`

Note: full `tests\test_api.py` currently has two unrelated failures in `TestCWE200ApiKeyLeak` because the class references `self.TEST_AGENT_ID`, which is not defined on that class.

## Summary by CodeRabbit

* **Bug Fixes**
  * Session validation now checks current server-side session state, rejecting tokens that are expired, missing, have mismatched session IDs, or are deactivated.
  * Deactivated session tokens are blocked from writing new memory data.
* **Tests**
  * Added an API contract test to confirm deactivated session tokens cannot write memory (request returns 401, and no document upload occurs).

### DIFF
--- a/memanto/app/services/session_service.py
+++ b/memanto/app/services/session_service.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from typing import Any
 
 import jwt
+from pydantic import ValidationError
 
 from memanto.app.config import get_data_dir, settings
 from memanto.app.core import create_memory_scope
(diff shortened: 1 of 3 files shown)


Xenogents (approved):

Xenogents:
Good catch! I don't think this is a critical enough issue to reward for the bounty but good work!

PR #815, opened 2026-06-25
fix(security): require API key for agent lifecycle routes (#770)

MyouzzZ:
/claim #770

## Summary

The V2 agent lifecycle routes were documented as requiring the server `MOORCHEH_API_KEY`, but the route dependency only checked that the server had a configured outbound key. As a result, unauthenticated callers could reach lifecycle operations such as creating/listing/deleting agents and activating an agent session.

This PR makes the request authentication match the documented contract: callers must provide the configured key via `Authorization: Bearer <MOORCHEH_API_KEY>` or `X-API-Key` before lifecycle routes can create agents, list agents, delete agents, activate sessions, deactivate sessions, or read `/api/v2/status`.

## Reproduction

The new regression tests capture the previously unsafe behavior:

- `test_create_agent_requires_authorization_header`: unauthenticated `POST /api/v2/agents` now returns `401` instead of creating an agent.
- `test_activate_session_requires_authorization_header`: unauthenticated `POST /api/v2/agents/{agent_id}/activate` now returns `401` and never returns a `session_token`.
- `test_create_agent_accepts_x_api_key_header`: verifies the documented `X-API-Key` alternative still works.

## Fix

- `verify_moorcheh_api_key` now extracts a request key from `Authorization: Bearer ...` or `X-API-Key` and compares it to the configured server key with `hmac.compare_digest`.
- `create_agent` now uses the same verifier as the other lifecycle endpoints.
- `GET /api/v2/status` is also protected because it exposes active session metadata.
- On-prem deployments fail closed unless an operator configures `MOORCHEH_API_KEY` as the local HTTP admin key.
- The existing `TestCWE200ApiKeyLeak` filename-sanitization tests now define the agent id constant they already referenced, so the full API test module passes.

## Validation

```bash
python -m pytest tests/test_api.py -q
# 57 passed

python -m pytest tests/test_unit.py -q
# 23 passed

python -m ruff check memanto/app/routes/auth_deps.py memanto/app/routes/sessions.py tests/test_api.py
# All checks passed
```

## Summary by CodeRabbit

* **New Features**
  * API-protected agent creation and status endpoints now require a Moorcheh API key using either `Authorization: Bearer <key>` or `X-API-Key`.

* **Bug Fixes**
  * Requests without a provided API key are rejected with a clear unauthorized response; mismatched keys are rejected with a forbidden response using constant-time comparison.

* **Tests**
  * Added coverage to ensure agent creation and session activation fail without auth headers, accept `X-API-Key`, and that global status requests include auth headers.

### DIFF
--- a/memanto/app/routes/auth_deps.py
+++ b/memanto/app/routes/auth_deps.py
@@ -4,6 +4,8 @@ Authentication Dependencies for V2 API
 Shared authentication utilities to avoid circular imports.
 """
 
+import hmac
+
 from fastapi import Header, HTTPException
 
 from memanto.app.models.session import Session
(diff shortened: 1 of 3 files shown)


Xenogents:
Thanks for your contributions! We're closing this PR as some of our recent security changes already fix this issue, sorry for getting back so late.

PR #823, opened 2026-06-25
Fix as-of memory expiry filtering

n4p68vc6p9-hub:
## Summary
- let `search_as_of` retrieve expired-now memories before applying point-in-time validity checks
- keep default TTL enforcement for current/recent memory fetch paths
- add a regression test for memories that were valid at the requested historical date but expired later

## Root cause
`search_as_of` calls `_fetch_all_memories`, and `_fetch_all_memories` always filtered expired memories against the current wall-clock time. That dropped memories that are expired today even when they were still valid at the requested `as_of_date`, so historical recall could lose the exact state it was supposed to reconstruct.

## Validation
- `python -m pytest tests/test_unit.py::TestMemoryReadServiceTemporal::test_search_as_of_includes_memory_valid_then_but_expired_now -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved time-based memory search so results are evaluated against the requested “as of” timestamp more accurately.
  * Fixed an issue where valid memories could be incorrectly excluded when their expiration time falls after the requested date.
* **Tests**
  * Added regression coverage for memory searches using historical timestamps.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -268,7 +268,9 @@ class MemoryReadService:
                     "temporal_mode": "as_of",
                 }
 
-            all_memories = self._fetch_all_memories(namespaces, type=type, tags=tags)
+            all_memories = self._fetch_all_memories(
+                namespaces, type=type, tags=tags, enforce_ttl=False
+            )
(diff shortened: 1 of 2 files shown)


het0814:

PR #825, opened 2026-06-25
Filter secrets from extracted memories

n4p68vc6p9-hub:
## Summary
- add deterministic secret-pattern filtering to conversation memory extraction
- skip extracted candidates that contain credential-like values in the title or content
- add regression coverage for a model response that mixes a credential candidate with a safe memory

## Root cause
The extraction prompt told the model not to return secrets, but `_normalize_candidates` trusted the model output completely. If the answer-generation step returned an API key, token, password, or similar credential anyway, the candidate could be persisted as long-term memory.

## Validation
- `python -m pytest tests/test_conversation_memory_extraction.py -q`
- `python -m pytest tests/test_api.py::TestMEMANTOAPI::test_extract_memories_from_conversation_dry_run tests/test_api.py::TestMEMANTOAPI::test_extract_memories_from_conversation_stores_batch -q`

### DIFF
--- a/memanto/app/services/conversation_memory_extraction_service.py
+++ b/memanto/app/services/conversation_memory_extraction_service.py
@@ -20,6 +20,16 @@ class ConversationMemoryExtractionService:
     MAX_MESSAGES = 200
     MAX_MEMORIES = 100
     MAX_CONTENT_CHARS = 12_000
+    SECRET_PATTERNS = [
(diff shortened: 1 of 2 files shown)


Xenogents:
After reviewing your PR, we've determined that this is an issue that would have to be resolved on backend side or right before all remember operations.

PR #828, opened 2026-06-25
Preserve memory metadata during updates

LAieh12:
/claim #770
Fixes #770

## Summary

MemoryWriteService.update_memory recreates a memory document after applying edits, but the reconstructed MemoryRecord only looked for the existing memory type in metadata["type"]. Stored Moorcheh documents use memory_type, and MemoryReadService formats that field as type. As a result, editing only the content/title of a non-fact memory could silently re-upload it as fact and reset provenance/trust metadata.

## Fix

- read existing fields from both formatted records and raw metadata without treating falsey values as missing
- preserve memory_type, provenance, validation_count, contradiction_detected, source_ref, tags, and TTL fields when they are not being updated
- add a regression test that edits only content on a preference memory and asserts the re-uploaded document remains a preference with trust metadata intact

## Validation

- uv run --group dev python -m pytest tests/test_unit.py -q
- uv run --group dev python -m pytest tests/test_api.py::TestMEMANTOAPI::test_edit_memory_with_session tests/test_api.py::TestMEMANTOAPI::test_edit_memory_rejects_empty_update tests/test_api.py::TestMEMANTOAPI::test_edit_memory_returns_404_when_missing -q
- uv run --group dev ruff check memanto/app/services/memory_write_service.py tests/test_unit.py

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory updates now preserve existing fields (including falsey trust/provenance/validation values) when the incoming update omits them.
  * Improved robustness when metadata is missing or not a dictionary.
  * TTL behavior now preserves existing `ttl_seconds`/`expires_at` when no new TTL is provided.
  * Memory read formatting now safely extracts nested metadata without dropping falsey values.
  * Document serialization now includes `ttl_seconds = 0` (but still omits when `None`).

* **Tests**
  * Added unit coverage for update behavior, including document payload composition and TTL omission/preservation rules.

### DIFF
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -115,7 +115,7 @@ class MemoryRecord(BaseModel):
             document["tags"] = ",".join(self.tags)  # Comma-separated for filtering
         if self.expires_at:
             document["expires_at"] = self.expires_at.isoformat()
-        if self.ttl_seconds:
+        if self.ttl_seconds is not None:
             document["ttl_seconds"] = self.ttl_seconds
         if self.superseded_by:
             document["superseded_by"] = self.superseded_by
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -746,13 +746,18 @@ class MemoryReadService:
 
         # Check if metadata is in nested format (Moorcheh API spec)
         metadata = item.get("metadata", {})
+        if not isinstance(metadata, dict):
+            metadata = {}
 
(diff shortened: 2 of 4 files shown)


Neelpatel1604:
Hey @LAieh12  - thanks for taking this on and for the work on #770. Really helpful catch on metadata getting dropped during updates.
We’re going to close this PR for now - it’s conflicted with current main and the update flow has changed since you opened it, so it’d need a fresh rebase rather than a straight merge.

PR #852, opened 2026-06-26
Preserve underscored scope ids in namespace parsing

n4p68vc6p9-hub:
## Summary
- preserve the full scope_id when parsing MEMANTO namespaces with underscores
- add a regression test for round-tripping an allowed underscored agent id

## Root Cause
`AgentCreate` allows underscores in agent identifiers, so `sales_bot` is valid and produces the namespace `memanto_agent_sales_bot`. `MemoryScope.from_namespace()` split on every underscore and required exactly three parts, causing valid namespaces with underscored scope IDs to be rejected as invalid.

## Tests
- `.venv\Scripts\python.exe -m pytest tests/test_unit.py::TestMEMANTOArchitecture::test_namespace_round_trip_preserves_underscored_scope_id -q`
- `.venv\Scripts\python.exe -m pytest tests/test_unit.py -q`
- `.venv\Scripts\python.exe -m py_compile memanto/app/core.py tests/test_unit.py`

Related to #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved namespace handling so identifiers containing underscores are preserved correctly.
  * Ensured namespace parsing still validates the expected format while supporting round-trip consistency.
* **Tests**
  * Added coverage for namespace round-tripping with underscore-containing identifiers.

### DIFF
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -34,7 +34,7 @@ class MemoryScope(BaseModel):
         """Parse namespace back to scope"""
         from typing import cast
 
-        parts = namespace.split("_")
+        parts = namespace.split("_", 2)
         if len(parts) != 3 or parts[0] != "memanto":
             raise ValueError(f"Invalid MEMANTO namespace format: {namespace}")
         return cls(scope_type=cast(ScopeType, parts[1]), scope_id=parts[2])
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #859, opened 2026-06-26
fix(migrate): make imported memory ids stable across retries

VoltVoks:
## Summary
- assign retry-stable UUIDv5 ids to mapped migration rows before batch import
- scope imported ids by target agent, provider, and source identity/content fingerprint
- pass caller-provided ids through both CLI clients so migration retries target the same records

## Flaw
A transient failure after an earlier migration chunk succeeds can leave imported memories committed. Retrying the same export previously submitted those already-imported rows without ids, so each retry generated fresh memory ids and created duplicate imported memories.

## Reproduction
1. Prepare a Mem0 export with more than 100 memories so migration runs in multiple batches.
2. Let the first batch succeed and force the second batch to raise a transient error.
3. Retry the same migration after the transient error is gone.
4. Before this change, the first 100 source records are submitted again without stable ids, so the write path assigns new ids and duplicates the already-imported memories.

## Fix
The migration runner now assigns deterministic ids before batching. Rows with a provider `source_ref` use that stable source identity; rows without one fall back to a content fingerprint. The id namespace includes the target agent and provider so imports remain retry-safe without colliding across agents. Both CLI clients now forward caller-provided ids into `MemoryRecord`.

## Tests
- python -m pytest tests/test_migrate_runner.py -q
- python -m ruff check tests/test_migrate_runner.py memanto/cli/migrate/runner.py memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py
- git diff --check

Refs #770

## Summary by CodeRabbit

* **New Features**
  * Batch imports can now preserve item IDs when provided, helping keep records consistent across writes and migrations.
  * Migration imports now generate stable, repeatable IDs for items that don’t already have one.

* **Bug Fixes**
  * Improved retry behavior during migrations so partially imported items keep the same ID on subsequent attempts.
  * Import IDs are now scoped per agent, reducing the chance of collisions between different sources.

### DIFF
--- a/memanto/cli/client/direct_client.py
+++ b/memanto/cli/client/direct_client.py
@@ -720,7 +720,7 @@ class DirectClient:
                 "source": item.get("source") or "user",
                 "provenance": provenance,
             }
-            for opt_key in ("source_ref", "created_at", "updated_at"):
+            for opt_key in ("id", "source_ref", "created_at", "updated_at"):
                 val = item.get(opt_key)
                 if val is not None:
                     kwargs[opt_key] = val
--- a/memanto/cli/client/sdk_client.py
+++ b/memanto/cli/client/sdk_client.py
@@ -549,7 +549,7 @@ class SdkClient:
                 "source": item.get("source") or "user",
                 "provenance": provenance,
             }
-            for opt_key in ("source_ref", "created_at", "updated_at"):
+            for opt_key in ("id", "source_ref", "created_at", "updated_at"):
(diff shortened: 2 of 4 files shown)


Xenogents:

Regarding the PR, our memanto edit workflow now works by upserting the memory with the same id, so a strict memory -> uuid relationship is not always maintained and this can cause edited memories to be overwritten. With that in mind we'll stick to generating random uuids and accept the loss of duplication on retries.

PR #860, opened 2026-06-26
fix(conflicts): resolve filtered UI entries by original index

VoltVoks:
## Summary
- include each unresolved conflict's original JSON-file index in `list_conflicts()` results
- make the web conflict UI resolve by that original index instead of the filtered display index
- add regression coverage for filtered conflict lists and resolve round-trips

## Flaw
The conflict list API only returned unresolved conflicts, so the web UI rendered a filtered array and sent that filtered array index back to `/api/ui/conflicts/resolve`. The resolver indexes into the full conflict report file. If an earlier conflict was already resolved, visible conflict `0` could refer to full-file conflict `1`, but the UI still sent `conflict_index: 0` and resolved the wrong record.

## Reproduction
1. Create a conflict report with conflict 0 already resolved and conflict 1 unresolved.
2. Open the web UI conflicts list for that agent/date.
3. The UI shows only the unresolved conflict as visible entry 0.
4. Click a resolution action.
5. Before this change, the backend received `conflict_index: 0`, so it marked or deleted memory for the already-resolved full-file entry instead of the visible unresolved conflict.

## Fix
`list_conflicts()` now returns unresolved conflicts with a `conflict_index` field that preserves each conflict's original full-report index. The web UI uses that value for card ids, manual-resolution ids, and resolve requests, falling back to the display index only for older data.

## Tests
- python -m pytest tests/test_conflict_resolution.py -q
- python -m pytest tests/test_conflict_resolution.py tests/test_api.py::TestMEMANTOAPI::test_conflicts_list_api tests/test_api.py::TestMEMANTOAPI::test_conflicts_resolve_api -q
- python -m ruff check tests/test_conflict_resolution.py memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py
- git diff --check

Refs #770

## Summary by CodeRabbit

* **New Features**
  * Conflict Resolution now shows stable conflict numbers, even when some conflicts are already resolved.
  * Actions and manual resolution inputs now stay aligned with the correct conflict entry.

* **Bug Fixes**
  * Improved conflict handling so resolving one item no longer shifts or mislabels others.
  * Conflict lists now preserve the original position of each unresolved item for more reliable resolution.

### DIFF
--- a/memanto/app/ui/static/index.html
+++ b/memanto/app/ui/static/index.html
@@ -4335,16 +4335,17 @@
 
                 conflicts.forEach((c, i) => {
                     const ctype = (c.type || 'conflict').toUpperCase();
+                    const conflictIndex = Number.isInteger(c.conflict_index) ? c.conflict_index : i;
                     const explanation = c.explanation || c.reason || 'Conflicting information detected.';
 
                     html += `
-            <div class="conflict-card" id="conflict-${i}">
+            <div class="conflict-card" id="conflict-${conflictIndex}">
                 <div class="conflict-header">
                     <div>
                         <span class="conflict-type">${escHtml(ctype)}</span>
                         <span class="badge badge-warning" style="margin-left:8px">Unresolved</span>
                     </div>
-                    <span style="font-size:11px;color:var(--text-dim);font-family:var(--font-mono)">#${i}</span>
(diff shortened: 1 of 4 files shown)


Neelpatel1604:
Thanks for the PR - please rebase onto latest main and resolve conflicts. We're consolidating related fixes and will review again once updated.

PR #862, opened 2026-06-26
fix(sync): refresh memory export before project sync

VoltVoks:
/claim #770

## Summary
- Fixes `memanto memory sync` in direct-client mode so it refreshes the export before copying `MEMORY.md` into a project.
- Aligns sync result metadata so refreshed exports are reported as `source: "fresh"` instead of implying a cached file was used.
- Adds regression coverage for stale cached exports and the CLI sync path.

## Reproduction
1. Create an agent export at `~/.memanto/exports/<agent>_memory.md`.
2. Add or update memories for the same agent.
3. Run `memanto memory sync --project-dir <project>` while using the direct client path.
4. Before this fix, the command copied the existing cached export without calling `export_memory_md()`, so `<project>/MEMORY.md` could silently lag behind current agent memory even though the CLI described sync as fresh.

## Fix
Direct sync now refreshes the export first, then copies the generated export to the target project. SDK sync already refreshed first, but its result metadata now reports that accurately.

## Validation
- `.venv/bin/python -m pytest tests/test_memory_sync.py tests/test_cli.py::TestMEMANTOCLI::test_memory_sync -q`
- `.venv/bin/python -m ruff check tests/test_memory_sync.py memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory sync now refreshes exported memory before copying it into a project, helping prevent outdated content in `MEMORY.md`.
  * Sync results now report a more accurate memory count when a fresh export is created, instead of always showing zero.
* **Tests**
  * Added coverage for refreshed memory syncing and correct result reporting.


Xenogents:
Your PR has been merged but your changes have been reverted. It's intended for the memory sync to used the cached export so it's not a hefty operation. Thanks for your contributions to Memanto!

PR #869, opened 2026-06-27
fix(conflicts): keep unresolved when deletion fails

Dreamstore2046:
## Summary
- stop conflict resolution from marking a report entry as resolved when required memory deletion fails
- apply the same behavior to both `DirectClient` and `SdkClient`
- add regression coverage for `keep_old` and `manual` resolutions after backend delete failures

## Flaw
`resolve_conflict()` catches `delete_memory()` failures as warnings, then still writes `resolved: true` and `resolution` back to the local conflict report. For actions such as `keep_old`, `keep_new`, `remove_both`, and `manual`, deletion is the operation that actually enforces the requested resolution.

If that delete fails, the underlying memory remains active while the conflict disappears from future `list_conflicts()` results because unresolved lists filter out entries with `resolved: true`. The `manual` path could also create a replacement memory while both conflicting originals remain.

## Reproduction
1. Create a conflict report with `old_memory_id` and `new_memory_id`.
2. Make the backend delete call fail, for example due to a transient backend error or unavailable API.
3. Call `resolve_conflict(..., action="keep_old")`.
4. Before this change, the JSON report was still marked `resolved: true`, hiding an unresolved conflict even though the new memory was not deleted.

## Fix
The clients now collect required deletion failures and raise `ValueError` before writing the report as resolved. The `manual` action also aborts before storing a replacement if required cleanup failed, so users do not get a third memory layered on top of unresolved originals.

## Tests
- `python -m pytest tests/test_conflict_resolution_delete_failures.py -q`
- `python -m pytest tests/test_conflict_resolution_delete_failures.py tests/test_api.py::TestMEMANTOAPI::test_conflicts_resolve_api -q`
- `python -m pytest tests/test_cli.py::TestMEMANTOCLI::test_conflicts_list -q`
- `python -m ruff check memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py tests/test_conflict_resolution_delete_failures.py`
- `python -m py_compile memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py tests/test_conflict_resolution_delete_failures.py`
- `git diff --check`

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Conflict resolution now fails fast when required memory cleanup doesn’t fully succeed, for both supported client modes.
  * If required deletions raise errors or report failure, conflicts are not marked as resolved and resolution details are not written.
  * Manual conflict resolution no longer stores replacement content when required deletions fail.

* **Tests**
  * Added coverage to confirm `keep_old` and `manual` actions raise an error and leave conflict reports unresolved when required deletions either raise or return unsuccessful results.

### DIFF
--- a/memanto/cli/client/direct_client.py
+++ b/memanto/cli/client/direct_client.py
@@ -1363,6 +1363,7 @@ class DirectClient:
 
         write_service = self._get_write_service()
         result_details = {"action": action}
+        delete_failures: list[str] = []
 
         if action == "keep_old":
             # Keep old, delete new
@@ -1371,7 +1372,7 @@ class DirectClient:
                     write_service.delete_memory(new_id, namespace)
                     result_details["deleted"] = new_id
                 except Exception as e:
-                    result_details["warning"] = f"Could not delete new memory: {e}"
+                    delete_failures.append(f"new memory {new_id}: {e}")
 
         elif action == "keep_new":
             # Keep new, delete old
@@ -1380,7 +1381,7 @@ class DirectClient:
                     write_service.delete_memory(old_id, namespace)
                     result_details["deleted"] = old_id
                 except Exception as e:
-                    result_details["warning"] = f"Could not delete old memory: {e}"
+                    delete_failures.append(f"old memory {old_id}: {e}")
 
         elif action == "keep_both":
             # No-op — both memories remain active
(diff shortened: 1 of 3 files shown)


Dreamstore2046:
@coderabbitai review

Dreamstore2046:
Rebased this PR onto the latest main and re-ran the focused validation.\n\nCurrent coverage includes the exception and falsey-delete failure paths, including the manual-resolution case where a failed required delete must block storing the replacement memory.\n\nValidation:\n- uv run --group dev pytest tests/test_conflict_resolution_delete_failures.py -q\n- uv run --group dev ruff check memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py tests/test_conflict_resolution_delete_failures.py\n- git diff --check

PR #1206, opened 2026-06-27
Fix namespace parsing for scope IDs containing underscores

2lll5:
## Summary
- Fix `MemoryScope.from_namespace` so namespace parsing preserves scope IDs that contain underscores
- Add a regression test covering round-trip conversion for `memanto_agent_customer_support_v2`

## Bug / impact
`MemoryScope.to_namespace()` emits `memanto_{scope_type}_{scope_id}` and `validate_namespace_format()` permits underscores in `scope_id`, but `from_namespace()` used `namespace.split("_")`. Any valid scope ID such as `customer_support_v2` produced more than three parts and raised `ValueError`, breaking namespace round-tripping and downstream retrieval/validation flows for common agent/session IDs.

## Verification
- `python3 -m pytest tests/test_unit.py tests/test_memory_parsing.py -q`
- `python3 -m ruff check memanto/app/core.py tests/test_unit.py`

Relates to #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed namespace parsing so scope IDs containing underscores are preserved correctly when converting to and from namespace strings.

* **Tests**
  * Added coverage to verify underscore-containing scope IDs round-trip without losing information.

### DIFF
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -34,7 +34,7 @@ class MemoryScope(BaseModel):
         """Parse namespace back to scope"""
         from typing import cast
 
-        parts = namespace.split("_")
+        parts = namespace.split("_", 2)
         if len(parts) != 3 or parts[0] != "memanto":
             raise ValueError(f"Invalid MEMANTO namespace format: {namespace}")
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #1209, opened 2026-06-28
fix: namespace parsing crash on underscore scope_ids + TTL timezone mismatch TypeError

clarboncy:
### Bug 1: `MemoryScope.from_namespace()` crashes when `scope_id` contains underscores

**Impact:** Any user/agent with an underscore in their ID (e.g. `john_doe`, `my_bot_v2`) triggers an unhandled `ValueError` during namespace roundtrip. This breaks memory retrieval for all such users.

**Root cause:** `namespace.split("_")` requires exactly 3 parts. `memanto_user_john_doe` splits into 4 parts → `ValueError`.

**Fix:** Use `split("_", maxsplit=1)` on the remainder after the `memanto_` prefix, preserving the full scope_id.

### Bug 2: `_filter_expired_memories()` raises `TypeError` on mixed timezone awareness

**Impact:** When `expires_at` is stored as a naive datetime (via `datetime.utcnow()` in `MemoryRecord`) but the filter compares against `datetime.now(timezone.utc)` (aware), Python raises `TypeError: can't compare offset-naive and offset-aware datetimes`. This crashes memory retrieval for any memory with a TTL.

**Root cause:** `MemoryRecord.created_at` uses `datetime.utcnow()` (naive) while `_filter_expired_memories` uses `datetime.now(timezone.utc)` (aware). Comparing them raises `TypeError`.

**Fix:** Check `tzinfo` and normalize both sides to the same awareness before comparison. Also added `TypeError` to the exception handler.

### Tests

13 regression tests in `tests/test_namespace_ttl_bugs.py`:
- 7 tests for namespace parsing (simple, underscore, multiple underscores, roundtrip, invalid cases)
- 6 tests for TTL filtering (naive/aware future/past, no expires_at, mixed naive+aware)

All tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved namespace handling so identifiers with underscores are parsed correctly and invalid formats are rejected more reliably.
  * Fixed memory expiration checks to work consistently with both timezone-aware and timezone-naive dates.
  * Reduced the chance of unexpected filtering errors when expiration values are missing or malformed.

* **Tests**
  * Added regression coverage for namespace parsing edge cases.
  * Added regression coverage for TTL expiration behavior across mixed datetime formats.

### DIFF
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -31,13 +31,21 @@ class MemoryScope(BaseModel):
 
     @classmethod
     def from_namespace(cls, namespace: str) -> "MemoryScope":
-        """Parse namespace back to scope"""
+        """Parse namespace back to scope.
+
+        Uses maxsplit=2 so that scope_id values containing underscores
+        are preserved correctly.  E.g. ``memanto_user_john_doe`` parses
+        to scope_type=``user``, scope_id=``john_doe``.
+        """
         from typing import cast
 
-        parts = namespace.split("_")
-        if len(parts) != 3 or parts[0] != "memanto":
+        if not namespace.startswith("memanto_"):
             raise ValueError(f"Invalid MEMANTO namespace format: {namespace}")
-        return cls(scope_type=cast(ScopeType, parts[1]), scope_id=parts[2])
+        remainder = namespace[len("memanto_"):]
+        parts = remainder.split("_", maxsplit=1)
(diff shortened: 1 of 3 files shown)


Xenogents:
Thanks for your contributions! We're closing this PR as these changes have already been fixed by other PR's.

PR #1210, opened 2026-06-28
Fix multi-scope recall import path

Dreamstore2046:
## Summary
- Fix `search_multi_scope()` so it imports `ScopeType` from the canonical `memanto.app.constants` module.
- Add a regression test that exercises multi-scope namespace construction and backend search invocation.

## Bug / impact
`search_multi_scope()` currently imports `ScopeType` from `memanto.memanto.app.constants`, which does not exist in the installed package layout. Any multi-scope recall path fails before it can build namespaces or query Moorcheh.

## Tests
- `.venv\Scripts\python.exe -m pytest tests/test_unit.py::TestMemoryReadServiceMultiScope -q`
- `.venv\Scripts\python.exe -m pytest tests/test_unit.py::TestMemoryWriteServiceDelete tests/test_unit.py::TestMemoryReadServiceMultiScope -q`
- `.venv\Scripts\python.exe -m pytest tests/test_unit.py -q`

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected how multi-scope memory search references the app’s constants, helping ensure searches use the intended namespaces.

* **Tests**
  * Added coverage for multi-scope memory search to verify the expected namespaces are sent, the search limit is applied, and the returned namespace list matches the query.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -175,7 +175,7 @@ class MemoryReadService:
             namespaces = []
             from typing import cast
 
-            from memanto.memanto.app.constants import ScopeType
+            from memanto.app.constants import ScopeType
 
             for scope_def in scopes:
                 scope = create_memory_scope(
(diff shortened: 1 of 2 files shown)


Dreamstore2046:
Closing this because it overlaps with existing PR #1201 for the same multi-scope import-path failure. Keeping the queue clean so the maintainer can review the non-duplicate fixes.

PR #1213, opened 2026-06-28
Guard conflict report parsing against non-object items

Dreamstore2046:
## Summary
- Preserve conflict report generation when the LLM returns a valid JSON array containing non-object items.
- Reuse the existing "Unparsed conflict report" fallback for malformed schemas as well as malformed JSON.
- Add a regression test for a legal JSON response like `["not an object", 1]`.

## Why
The conflict report parser assumed every parsed array item was a dictionary. If the model returned a valid JSON array with strings or numbers, report generation crashed with `AttributeError` before saving the raw response for review.

Refs #770

## Validation
- `python -m pytest tests/test_daily_analysis_service.py tests/test_unit.py -q`
- `python -m ruff check memanto/app/services/daily_analysis_service.py tests/test_daily_analysis_service.py`
- `python -m py_compile memanto/app/services/daily_analysis_service.py tests/test_daily_analysis_service.py`
- `git diff --check`

## Xenogents Changes
- Combine isolated test file with other tests
- Improve malformed json fallback logic

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conflict report handling when AI output is malformed, preventing crashes on unexpected JSON shapes.
  * Invalid AI responses now fall back to a single unparsed conflict entry when appropriate, preserving the original content for review.

* **Tests**
  * Added coverage for malformed conflict items to confirm the report still generates successfully and keeps the raw payload visible.


Xenogents (approved):

Xenogents:
Thanks for your work! This is a good change though not a big enough issue to be considered for the bounty, thanks for your contributions to Memanto!

PR #1214, opened 2026-06-28
Isolate MCP client sessions per agent

SueJianjian:
/claim #770

## Summary

Fixes an MCP session isolation bug where concurrent tool calls for different `agent_id`s could share and overwrite the same `SdkClient` session state.

`SdkClient` keeps the active `agent_id` and `session_token` on the instance. Before this change, `MemantoLifecycle` had one shared client. `ensure_ready()` was locked only while activating the session, then tool handlers called `lifecycle.client.remember()` / `recall()` outside the lock. If another MCP call activated a different agent in between, the first call could run with the wrong active client state and fail with a session mismatch.

## Fix

- Keep a separate `SdkClient` per agent inside `MemantoLifecycle`.
- Return the ready, agent-scoped client from `ensure_ready()`.
- Cache newly created scoped clients only after agent readiness succeeds.
- Update MCP memory tools to use that scoped client for session-backed operations.
- Validate batch payloads before creating/activating an agent session.
- Keep admin tools on a separate admin client because they are not bound to one active session.
- Add regression coverage proving distinct clients, client reuse, missing-agent auto-create via scoped client, failed activation retry behavior, and side-effect-free invalid batch validation.

## Validation

```text
PYTHONPATH=<repo>;<repo>/integrations/mcp python -m pytest integrations/mcp/tests/test_lifecycle.py integrations/mcp/tests/test_config.py integrations/mcp/tests/test_server.py -q
26 passed

python -m py_compile integrations/mcp/memanto_mcp/lifecycle.py integrations/mcp/memanto_mcp/tools.py integrations/mcp/tests/test_lifecycle.py
```

Refs #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved agent-specific behavior in MCP by separating agent-management and agent-operations connections.
  * MCP tools now resolve the agent id once and perform remember/recall/answer through a ready, agent-scoped client for more consistent execution.
* **Tests**
  * Added integration-style tests verifying distinct per-agent scoped clients, reuse on subsequent readiness calls, auto-creation/activation for missing agents, correct caching after failed activation, and input validation happening before any lifecycle side effects.

### DIFF
--- a/integrations/mcp/memanto_mcp/lifecycle.py
+++ b/integrations/mcp/memanto_mcp/lifecycle.py
@@ -87,9 +87,9 @@ class MemantoLifecycle:
         """
         with self._lock:
             client = self._clients.get(agent_id)
-            if client is None:
+            is_new_client = client is None
+            if is_new_client:
                 client = SdkClient(api_key=self._settings.api_key_value())
-                self._clients[agent_id] = client
 
             if agent_id not in self._ensured_agents:
                 self._ensure_agent_exists_locked(client, agent_id)
@@ -101,6 +101,9 @@ class MemantoLifecycle:
             ):
                 self._activate_locked(client, agent_id)
                 self._activated_agents.add(agent_id)
+
+            if is_new_client:
+                self._clients[agent_id] = client
 
         return client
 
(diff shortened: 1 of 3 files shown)


Xenogents:
Thanks for your work! We'll be considering your contributions for when we determine a bounty winner.

PR #1215, opened 2026-06-28
Fix changed-since sorting with null updated_at

Dreamstore2046:
## Summary
- Make `search_changed_since()` sort changed memories by `updated_at` with a `created_at` fallback.
- Parse sort timestamps into comparable datetimes so `updated_at: null` cannot crash mixed result sets.
- Add a regression test for a newly created memory with `updated_at=None` returned alongside an updated memory.

## Root cause / impact
`search_changed_since()` correctly includes newly created memories when `created_at` is after the requested boundary, even when `updated_at` is null. It then sorted results with `m.get("updated_at", m.get("created_at", ""))`, which returns `None` when the `updated_at` key exists with a null value. Mixed `None` and string sort keys raise `TypeError`, so the changed-since recall path can fail instead of returning valid temporal results.

## Validation
- `..\memanto\.venv\Scripts\python.exe -m pytest tests\test_unit.py::TestMemoryReadServiceChangedSince::test_changed_since_sorts_created_memories_without_updated_at -q`
- `..\memanto\.venv\Scripts\python.exe -m pytest tests\test_unit.py -q`
- `..\memanto\.venv\Scripts\python.exe -m pytest tests\test_api.py::TestMEMANTOAPI::test_recall_temporal_api -q`
- `..\memanto\.venv\Scripts\python.exe -m py_compile memanto\app\services\memory_read_service.py tests\test_unit.py`
- `..\memanto\.venv\Scripts\python.exe -m ruff check memanto\app\services\memory_read_service.py tests\test_unit.py`
- `git diff --check`

Bounty: #770
/claim #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved “changed since” results ordering to consistently rank memories by the most recent change time, falling back to creation time when an update timestamp is missing or cannot be interpreted.
* **Tests**
  * Added a regression test to confirm correct sorting for memories without an update timestamp and verifies the expected change classification.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -16,12 +16,15 @@ from memanto.app.utils.errors import MemoryError
 
 
 class MemoryReadService:
+    """Read, search, and format memories from the configured Moorcheh backend."""
+
     def __init__(self, moorcheh_client: "MoorchehClient"):
         self.client = moorcheh_client
         self._namespace_service = None
 
     @property
     def namespace_service(self):
+        """Return the namespace service, creating it on first access."""
         if self._namespace_service is None:
             from memanto.app.services.namespace_service import NamespaceService
 
@@ -338,6 +341,7 @@ class MemoryReadService:
 
             # Sort by created_at descending (most recent first)
             def _created_sort_key(m: dict[str, Any]) -> str:
+                """Return a comparable created-at timestamp for recent ordering."""
                 raw = m.get("created_at")
                 if not raw:
                     return ""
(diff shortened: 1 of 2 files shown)


Dreamstore2046:
Rebased this PR onto the latest main and re-ran the focused validation.\n\nValidation:\n- uv run --group dev pytest tests/test_unit.py -q\n- uv run --group dev ruff check memanto/app/services/memory_read_service.py tests/test_unit.py\n- uv run --group dev python -m py_compile memanto/app/services/memory_read_service.py tests/test_unit.py\n- git diff --check

PR #1217, opened 2026-06-28
Skip malformed timestamps in temporal filtering

Dreamstore2046:
## Summary

- Fix temporal post-filtering so malformed per-memory `created_at` values are skipped instead of disabling the entire `created_after` / `created_before` filter.
- Keep the existing behavior for invalid caller-supplied temporal bounds: if the bound cannot be parsed, the filter is not applied.
- Add unit coverage for both lower-bound and upper-bound filtering with malformed result timestamps.

## Root cause

`_apply_temporal_filter()` wrapped the whole list comprehension in one `try` block. If any returned memory had an invalid `created_at`, `parse_iso_timestamp()` raised and the method fell through to `pass`, returning the unfiltered result set. That lets stale or future memories leak into temporal searches whenever one bad timestamp is present.

## Validation

- `python -m pytest tests/test_unit.py::TestMemoryReadServiceTemporalFilter -q`
- `python -m py_compile memanto/app/services/memory_read_service.py tests/test_unit.py`
- `git diff --check`

Bounty: #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved date filtering for memory results so only records with valid timestamps are included in time-based searches.
  * Records with missing or malformed timestamps are now excluded when applying date range filters, avoiding inconsistent results.

* **Tests**
  * Added coverage for time filtering behavior with invalid or missing timestamps to confirm the filtering rules work as expected.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -568,14 +568,32 @@ class MemoryReadService:
 
         filtered = results
 
+        def _created_at_matches(
+            result: dict[str, Any],
+            *,
+            after_dt=None,
+            before_dt=None,
+        ) -> bool:
+            created_at = result.get("created_at")
+            if not created_at:
+                return False
+            try:
+                created_dt = parse_iso_timestamp(created_at)
+            except (ValueError, AttributeError):
+                return False
+            if after_dt is not None and created_dt < after_dt:
+                return False
+            if before_dt is not None and created_dt > before_dt:
+                return False
+            return True
+
         if created_after:
             try:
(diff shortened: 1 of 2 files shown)


Dreamstore2046:
@coderabbitai review

Dreamstore2046:
Quick follow-up: the remaining CodeRabbit note about normalizing non-string `created_at` before parsing is already covered by the latest head commit `f891c0e`.

Current code uses `parse_iso_timestamp(str(created_at))` inside `_created_at_matches`, and the focused temporal-filter regression coverage includes a numeric `created_at` case. CodeRabbit status is now `SUCCESS`.

PR #811, opened 2026-06-25
Preserve falsey memory metadata values

n4p68vc6p9-hub:
## Summary
- Preserve explicit nested metadata values even when they are falsey, such as confidence=0.0.
- Treat non-dict metadata defensively before formatting memory results.
- Add a regression test so low-confidence memory metadata does not get converted to None.

Related to #770.

## Tests
- python -m pytest tests/test_unit.py::TestMemoryReadServiceFormatting::test_format_memory_item_preserves_falsey_metadata_values -q
- python -m pytest tests/test_unit.py -q

PR #1226, opened 2026-06-28
Preserve falsey nested metadata fields

Dreamstore2046:
Summary:
- Preserve explicit falsey values from nested Moorcheh metadata when formatting memory items.
- Only fall back to flat fields when the nested metadata key is absent, not when its value is `0`, `False`, or `[]`.
- Add regression coverage for `confidence: 0.0`, empty tags, zero validation count, false contradiction flags, and zero TTL values.

Reproduction from current service layer:
```py
from unittest.mock import MagicMock
from memanto.app.services.memory_read_service import MemoryReadService

item = {
    "text": "[FACT] Zero confidence\n\nStored with explicit falsey values.",
    "metadata": {"confidence": 0.0, "tags": [], "validation_count": 0, "contradiction_detected": False},
    "confidence": 0.95,
    "tags": "stale",
    "validation_count": 7,
    "contradiction_detected": True,
}
print(MemoryReadService(MagicMock())._format_memory_item(item))
```
Before this fix, nested `0.0` / `[]` / `0` / `False` values are treated as missing and overwritten by the flat fallback fields.

Validation:
- `uv run --group dev pytest tests\\test_unit.py::TestMemoryReadServiceFormat::test_format_memory_item_preserves_falsey_metadata_values -q`
- `uv run --group dev pytest tests\\test_unit.py -q`
- `uv run --group dev ruff check memanto\\app\\services\\memory_read_service.py tests\\test_unit.py`
- `uv run --group dev python -m py_compile memanto\\app\\services\\memory_read_service.py tests\\test_unit.py`

Parent bounty: #770

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -746,13 +746,17 @@ class MemoryReadService:
 
         # Check if metadata is in nested format (Moorcheh API spec)
         metadata = item.get("metadata", {})
+        if not isinstance(metadata, dict):
+            metadata = {}
 
         # Helper to get field from either nested metadata or flat structure
         def get_field(field_name, flat_field_name=None):
             """Get field from metadata object or fallback to flat field"""
             flat_name = flat_field_name or field_name
             # Try metadata object first (API spec), then flat field (fallback)
-            return metadata.get(field_name) or item.get(flat_name)
+            if field_name in metadata:
+                return metadata[field_name]
+            return item.get(flat_name)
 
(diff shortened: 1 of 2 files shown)


Dreamstore2046:
Closing this as a duplicate after rechecking the bounty thread: PR #811 already covers the same falsey nested metadata fallback issue and is still open. I missed that earlier claim because my duplicate search only checked current search terms, not the full bounty comment history. Keeping the thread clean and focusing on non-duplicative fixes.

PR #853, opened 2026-06-26
Normalize session datetimes before comparisons

n4p68vc6p9-hub:
## Summary
- normalize timezone-aware session datetimes before comparing or subtracting them
- apply the normalization to session status helpers, JWT validation, and ended-session duration calculation
- add regression tests for `Z`/timezone-aware session timestamps

## Root Cause
Session JSON and JWT payloads can contain ISO timestamps with a UTC timezone, such as `2026-03-19T20:00:00Z`. Pydantic parses those values as timezone-aware datetimes, but MEMANTO compares them with `utc_now()`, which intentionally returns a naive UTC datetime for legacy session storage. That raises `TypeError` in active-session checks and time-remaining calculations.

Reproduction on `main`:
```python
Session(
    session_id="sess-test",
    session_token="token-test",
    agent_id="agent",
    namespace="memanto_agent_agent",
    started_at="2026-03-19T14:00:00Z",
    expires_at="2099-03-19T20:00:00Z",
).is_expired()
```
This fails with `TypeError: can't compare offset-naive and offset-aware datetimes`.

## Tests
- `.venv\Scripts\python.exe -m pytest tests/test_unit.py::TestSessionService::test_session_status_handles_aware_expiration_timestamp tests/test_unit.py::TestSessionService::test_validate_session_handles_aware_expiration_timestamp -q`
- `.venv\Scripts\python.exe -m pytest tests/test_unit.py -q`
- `.venv\Scripts\python.exe -m pytest tests/test_cli.py -q`
- `.venv\Scripts\python.exe -m py_compile memanto/app/utils/temporal_helpers.py memanto/app/models/session.py memanto/app/services/session_service.py tests/test_unit.py`
- `git diff --check`

Related to #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session expiration and remaining-time calculations to handle timezone-aware timestamps consistently.
  * Session validation now correctly compares dates even when stored values include UTC offsets.

* **Tests**
  * Added coverage for timezone-aware session expiry, activity checks, and token validation behavior.

PR #1228, opened 2026-06-28
Keep session info stable with timezone-aware expiry

Dreamstore2046:
## Summary

Fixes a narrow legacy CLI session-status bug in `memanto session info`: JWT payloads can contain ISO expirations with a UTC offset such as `2026-03-19T20:00:00+00:00`, but the command compared that aware datetime against the existing naive UTC `datetime.utcnow()` clock. That raised `TypeError` internally and the command showed `Unknown (Token unreadable)` even though the token decoded correctly.

This normalizes aware JWT expiry values to naive UTC before computing status and time remaining. The change is scoped to the legacy CLI command and does not alter session validation or persistence.

## Why

The broader session datetime fix in #853 covers model/service comparisons, but `memanto/cli/commands/session.py` performs its own JWT decode and time comparison. Users can still see an unreadable/unknown active session in the legacy CLI alias when the token expiry includes `+00:00`.

## Validation

- `uv run --group dev pytest tests\test_cli.py::TestMEMANTOCLI::test_session_info tests\test_cli.py::TestMEMANTOCLI::test_session_info_handles_timezone_aware_token -q`
- `uv run --group dev pytest tests\test_cli.py -q`
- `uv run --group dev ruff check memanto\cli\commands\session.py tests\test_cli.py`
- `uv run --group dev python -m py_compile memanto\cli\commands\session.py tests\test_cli.py`

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session expiration handling so time remaining is calculated correctly for tokens with timezone-aware expiry timestamps.
  * Session details now display more reliably for active sessions instead of showing an unknown status in this case.

* **Tests**
  * Added coverage for viewing session info when the active token includes a timezone-aware expiration time.

### DIFF
--- a/memanto/cli/commands/session.py
+++ b/memanto/cli/commands/session.py
@@ -2,7 +2,7 @@
 MEMANTO CLI - Legacy session aliases for agent activation commands.
 """
 
-from datetime import datetime
+from datetime import datetime, timezone
 
 import jwt
 from rich.table import Table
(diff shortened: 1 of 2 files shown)


Dreamstore2046:
Rebased this PR onto the latest main and re-ran the focused validation.\n\nValidation:\n- uv run --group dev pytest tests/test_cli.py::TestMEMANTOCLI::test_session_info tests/test_cli.py::TestMEMANTOCLI::test_session_info_handles_timezone_aware_token -q\n- uv run --group dev pytest tests/test_cli.py -q\n- uv run --group dev ruff check memanto/cli/commands/session.py tests/test_cli.py\n- uv run --group dev python -m py_compile memanto/cli/commands/session.py tests/test_cli.py\n- git diff --check

PR #1236, opened 2026-06-28
Skip malformed memories in agent bootstrap

Dreamstore2046:
## Summary

Bounty issue: #770

This PR keeps `memanto agent bootstrap` stable when recall returns malformed memory items inside an otherwise successful response.

The bootstrap command fetches several recall samples, then deduplicates and renders them. It previously cast `result["memories"]` to `list[dict]` without validating the shape, so a non-object item such as a truncated string raised `AttributeError: 'str' object has no attribute 'get'` before any valid memories could be shown.

## Changes

- Normalize recall responses at the bootstrap fetch boundary.
- Skip non-object memory entries instead of letting them reach dedupe/rendering code.
- Add a regression test proving a malformed item does not suppress a valid memory.

This is separate from invalid confidence formatting: this bug is about the memory record itself not being an object.

## Validation

- `uv run --group dev pytest tests\test_cli.py::TestMEMANTOCLI::test_agent_bootstrap_skips_non_object_recall_memories -q`
- `uv run --group dev pytest tests\test_cli.py -q`
- `uv run --group dev ruff check memanto\cli\commands\agent.py tests\test_cli.py`
- `uv run --group dev python -m py_compile memanto\cli\commands\agent.py tests\test_cli.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the agent bootstrap command to handle unexpected memory response shapes more safely.
  * Non-object memory entries are now ignored, helping the command continue working when recall results include malformed items.
  * Valid memory items still display normally, and the command remains successful.
* **Tests**
  * Added coverage for agent bootstrap behavior when recall results contain invalid entries.

### DIFF
--- a/memanto/cli/commands/agent.py
+++ b/memanto/cli/commands/agent.py
@@ -7,7 +7,7 @@ import time
 from collections import Counter
 from datetime import datetime
 from pathlib import Path
-from typing import Any, cast
+from typing import Any
 
 import typer
 from rich.panel import Panel
@@ -294,6 +294,12 @@ def agent_bootstrap(
     console.print()
 
     # Helper: fetch memories safely
+    def _memory_records(value: Any) -> list[dict[str, Any]]:
+        """Return only object-shaped memories from a recall response."""
+        if not isinstance(value, list):
+            return []
+        return [item for item in value if isinstance(item, dict)]
+
     def _fetch(
         query: str, type: list[str] | None = None, limit: int = 10
     ) -> list[dict[str, Any]]:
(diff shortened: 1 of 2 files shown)


Xenogents:
Thanks for you work on this! We think it's better design to have a fail-fast approach where if we receive a malformed json we want it to fail so the issue can be caught and fixed instead of adding guards to hide the failure. We've closed many of your other PR's under that same mindset. Also for the future could you make similar changes all into a single PR? This adds a lot of PR bloat to the repo and it's fine to have multiple changes in the same PR.

PR #1237, opened 2026-06-28
Skip malformed recall result memories

Dreamstore2046:
Summary: Filter recall response entries so only object-shaped memories are rendered; non-object items no longer crash the CLI display. Added regression coverage for mixed malformed recall results. Tests: uv run --group dev pytest tests\test_cli.py -q -k recall_skips_non_object_memory_records; uv run --group dev pytest tests\test_cli.py -q; uv run --group dev ruff check memanto\cli\commands\memory.py tests\test_cli.py; uv run python -m py_compile memanto\cli\commands\memory.py tests\test_cli.py; git diff --check.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recall results handling so only valid memory entries are shown.
  * Prevented malformed or non-object items from appearing in the CLI output.
  * Added test coverage for mixed valid and invalid memory records to confirm correct display behavior.

### DIFF
--- a/memanto/cli/commands/memory.py
+++ b/memanto/cli/commands/memory.py
@@ -28,6 +28,13 @@ from memanto.cli.commands._shared import (
 )
 
 
+def _memory_records(value):
+    """Return only displayable memory records from a recall response."""
+    if not isinstance(value, list):
+        return []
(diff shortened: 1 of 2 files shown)


Dreamstore2046:
@coderabbitai review

PR #1238, opened 2026-06-28
Guard answer context memory display

Dreamstore2046:
Summary: Harden the user-facing memanto answer context display so malformed non-object context_memories entries are skipped and context scores are rendered from finite numeric values only. This is separate from recall result rendering because it applies to the RAG answer output path. Tests: uv run --group dev pytest tests\test_cli.py -q -k answer_skips_malformed_context_memories; uv run --group dev pytest tests\test_cli.py -q; uv run --group dev ruff check memanto\cli\commands\memory.py tests\test_cli.py; uv run python -m py_compile memanto\cli\commands\memory.py tests\test_cli.py; git diff --check.

### DIFF
--- a/memanto/cli/commands/memory.py
+++ b/memanto/cli/commands/memory.py
@@ -4,6 +4,7 @@ detect-conflicts, conflicts).
 """
 
 import json
+import math
 import sys
 import time
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


Dreamstore2046:
@coderabbitai review

PR #1245, opened 2026-06-28
fix: re-enable memory validation service (#770)

Votienduong2208:
## Summary

Re-enables the  which was completely bypassed with the comment `# skip validation for speed`. The validation was disabled in both `store_memory()` and `batch_store_memories()`, meaning **all** memory storage bypassed poisoning checks — including prompt injections, data leaks, and contradictory data.

### 1. `memanto/app/services/memory_write_service.py`
- Imported `MemoryValidationService` from legacy module
- Added lazy-initialized `validation_service` property
- Re-enabled validation call in `store_memory()` — previously commented out with `# skip validation for speed`
- Re-enabled validation call in `batch_store_memories()` — same bypass

### 2. `memanto/app/routes/memory.py`
- Imported rate limiter functions from `utils.rate_limiting` (prepares for follow-up wiring)
- Added rate limit enforcement in the `remember` endpoint

## Impact

Before this fix, `POST /api/v2/agents/{id}/remember` could accept arbitrary content without validation. The validation service checks:
- Critical memory types (fact, preference) require user confirmation, repetition, or tool grounding
- Prevents poisoning attacks via the memory layer
- Enforces trust boundaries for different provenance types

Closes aspect of #770 — Memanto Bug & Exploit Challenge

## Summary by CodeRabbit

* **New Features**
  * Memory submissions can now include an explicit `user_confirmed` signal (single and batch), which affects how validation treats the content.
  * Validation can route stored memories as **provisional** when fully confirmed conditions aren’t met.
* **Bug Fixes**
  * Remember and batch-remember now enforce session-scoped rate limiting.
  * Updated how validation results determine API `status` and when session summary logging occurs.
* **Tests**
  * Added/expanded contract and service tests to verify `user_confirmed` propagation, rate-limit enforcement, and correct provisional/rejected behaviors.


Neelpatel1604 (changes requested):
Thanks for tackling #770  - re-enabling validation is the right direction.

Request changes:

- Don’t overwrite validation_result after validate_memory(); honor action and skip upload on reject.
- Move ValidationPolicy out of legacy  it was removed from core.py and the legacy import fails today.
- Fix enforce_write_rate_limit(agent_id) (one arg) and apply to batch-remember too.
- Pass validation context (user_confirmed on request or session) if you want confirmed facts to store as active.

Happy to re-review once those are addressed.

Votienduong2208:
Addressed the requested changes in f66020c:\n\n- Preserved MemoryValidationService action/reason in single and batch writes.\n- Skip Moorcheh upload for reject/quarantine validation outcomes.\n- Moved validation service/policy to memanto.app.services with a legacy shim for compatibility.\n- Applied write rate limiting to batch-remember too.\n- Added user_confirmed request context for remember and batch-remember.\n- Added regression tests for validation outcomes, skipped uploads, rate limiting, and validation context.\n\nValidation run locally:\n- ruff check targeted files: pass\n- pytest -q: pass (live E2E tests skipped without real MOORCHEH_API_KEY)

Votienduong2208:
Addressed the follow-up requested changes in 4b1f3ed:\n\n- Rejected/quarantined validation outcomes are no longer reported as queued or logged into local session summaries.\n- Batch remember now passes per-item user_confirmed validation contexts so mixed confirmed/unconfirmed batches keep their individual state.\n- ValidationPolicy now uses settings.REQUIRE_VALIDATION_FOR as the source of truth for critical memory types.\n- Repetition count is computed via _check_repetition() when callers do not provide it, so the repeated-content validation path is reachable.\n- Added regression coverage for single-route rate limiting, rejected-route behavior, per-item batch context, validator context forwarding, and repetition-count handling.\n\nValidation run locally:\n- python -m py_compile targeted files: pass\n- git diff --check: pass\n- python -m ruff check targeted files: pass\n- python -m pytest tests/test_memory_write_service.py tests/test_api.py -q: pass\n- python -m pytest -q: pass (E2E tests skipped because no real MOORCHEH_API_KEY in local env)

Votienduong2208:
Addressed the CodeRabbit review feedback in commit 17ea582.\n\nSummary:\n- Updated batch accepted_ids selection to rely on accepted actions and exclude only explicit failure statuses.\n- Added regression coverage for accepted non-failure backend statuses while still skipping rejected writes.\n\nValidation:\n- python -m pytest tests/test_api.py -q ✅

Votienduong2208:
Addressed the latest CodeRabbit follow-up in b31ca0b.

Summary:
- Single `/remember` now treats accepted validation actions as locally accepted only when the backend status is not an explicit failure (`failed`, `rejected`, `quarantined`).
- Backend failed `store` / `store_provisional` results now keep their failed status instead of being rewritten to `queued`.
- Failed backend writes are no longer logged into the local session summary.
- Added regression coverage for `action: store` + `status: failed` on the single remember route.
- Updated the memory write service test helper with the required `agent_id` field so the full suite reflects the current `MemoryRecord` schema.

Validation run locally:
- `python -m py_compile memanto/app/routes/memory.py tests/test_api.py tests/test_memory_write_service.py` ✅
- `git diff --check` ✅
- `python -m ruff check memanto/app/routes/memory.py tests/test_api.py tests/test_memory_write_service.py` ✅
- `python -m pytest tests/test_memory_write_service.py tests/test_api.py -q` ✅
- `python -m pytest -q` ✅ (live E2E tests skipped locally because no real `MOORCHEH_API_KEY` is configured)

Xenogents:
Thanks for your contributions! Sorry we'll have to close this PR as we're realizing now that these fixes have already been addressed by earlier PR's.

PR #1251, opened 2026-06-28
Skip malformed Hermes recall hits

Dreamstore2046:
## Summary
- Filter Hermes recall results to object-shaped memory entries before formatting them into prefetch context.
- Apply the same guard to the Hermes `memanto_recall` tool response formatting.
- Add regression coverage for malformed recall rows in both Hermes consumption paths.

## Root cause
The Hermes provider assumed every recall hit returned by Memanto was a dict-like memory record. If the backend/client response contained one malformed non-object row, such as a truncated string, the prefetch formatter raised `AttributeError`, and the explicit `memanto_recall` tool returned an error instead of preserving valid memories.

This is separate from the existing Hermes prompt-boundary/tag delimiter fix: this PR handles malformed recall response item shapes, not memory text containing wrapper tags.

## Validation
- Pre-fix regression: `uv run --group dev pytest integrations/hermes-agents/tests/test_provider.py::test_format_recall_block_skips_malformed_entries integrations/hermes-agents/tests/test_provider.py::test_recall_tool_skips_malformed_results -q` failed with the prefetch `AttributeError` and a recall-tool error response.
- `uv run --group dev pytest integrations/hermes-agents/tests/test_provider.py -q`
- `uv run --group dev ruff check integrations/hermes-agents/hermes_memanto/provider.py integrations/hermes-agents/tests/test_provider.py`
- `uv run --group dev ruff format --check integrations/hermes-agents/hermes_memanto/provider.py integrations/hermes-agents/tests/test_provider.py`
- `uv run --group dev python -m py_compile integrations/hermes-agents/hermes_memanto/provider.py integrations/hermes-agents/tests/test_provider.py`
- `git diff --check`

### DIFF
--- a/integrations/hermes-agents/hermes_memanto/provider.py
+++ b/integrations/hermes-agents/hermes_memanto/provider.py
@@ -222,7 +222,7 @@ def _is_trivial_message(text: str) -> bool:
     return bool(_TRIVIAL_RE.match((text or "").strip()))
 
 
-def _memory_score(raw: dict) -> float | None:
+def _memory_score(raw: dict[str, Any]) -> float | None:
     """Coalesce the score key Memanto uses across endpoints."""
     score = raw.get("score")
     if score is None:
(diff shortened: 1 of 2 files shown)


Dreamstore2046:
@coderabbitai review

PR #1254, opened 2026-06-28
fix(crewai): harden recall result rendering

VoltVoks:
/claim #770

## Summary

This fixes a CrewAI integration recall rendering bug in `MemantoRecallTool`.

CrewAI recall assumed every item returned from `client.recall()["memories"]` was an object and assumed `tags` was already a list. If a backend/client response contained one malformed non-object row, the tool raised `AttributeError` and hid otherwise valid memories from the CrewAI agent. If tags arrived as a comma-separated string, matching shapes handled elsewhere in Memanto, CrewAI rendered character-level tags like `u, r, g...` instead of semantic tags.

The fix filters recall output to object-shaped memories before rendering and normalizes tags from either a list or comma-separated string.

## Reproduction

Added regression coverage proving:

- a malformed non-object memory row is skipped while valid memories are still returned
- comma-separated string tags render as `urgent, client`, not character-level tags

## Duplicate boundary

This is scoped to `integrations/crewai`. It is separate from the existing MCP, LangGraph, Hermes, CLI recall display, memory export, and bootstrap robustness submissions because CrewAI has its own tool wrapper and rendering path.

## Validation

- `uv run --with pytest --with pytest-asyncio --with pytest-timeout --with-editable . --with-editable integrations/crewai pytest integrations/crewai/tests/test_tools.py::test_memanto_recall_tool_skips_malformed_memory_rows integrations/crewai/tests/test_tools.py::test_memanto_recall_tool_normalizes_string_tags -q`
- `uv run --with pytest --with pytest-asyncio --with pytest-timeout --with-editable . --with-editable integrations/crewai pytest integrations/crewai/tests/test_tools.py -q`
- `uv run ruff check integrations/crewai/crewai_memanto/tools.py integrations/crewai/tests/test_tools.py`
- `uv run python -m py_compile integrations/crewai/crewai_memanto/tools.py integrations/crewai/tests/test_tools.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recall results handling so invalid/non-conforming memory entries are ignored instead of being shown.
  * Standardized how memory tags are rendered, whether tags are provided as a list or as a comma-delimited string.
* **Tests**
  * Added unit test coverage to ensure malformed recall rows are skipped.
  * Added unit test coverage to confirm tags normalize and display correctly.

### DIFF
--- a/integrations/crewai/crewai_memanto/tools.py
+++ b/integrations/crewai/crewai_memanto/tools.py
@@ -37,6 +37,22 @@ VALID_MEMORY_TYPES = (
 )
 
 
+def _normalize_tags(raw_tags: Any) -> list[str]:
+    """Return semantic tag strings from common Memanto tag shapes."""
+    if isinstance(raw_tags, str):
+        return [tag.strip() for tag in raw_tags.split(",") if tag.strip()]
+    if isinstance(raw_tags, list):
+        return [str(tag).strip() for tag in raw_tags if str(tag).strip()]
+    return []
+
+
+def _valid_recall_memories(raw_memories: Any) -> list[dict[str, Any]]:
+    """Filter recall output to object-shaped memories CrewAI can render."""
+    if not isinstance(raw_memories, list):
+        return []
+    return [mem for mem in raw_memories if isinstance(mem, dict)]
+
+
 class MemantoSetup:
     """
     Manages Memanto agent lifecycle for CrewAI integration.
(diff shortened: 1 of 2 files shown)


Xenogents:
Thanks for you work on this! We think it's better design to have a fail-fast approach where if we receive a malformed json we want it to fail so the issue can be caught and fixed instead of adding guards to hide the failure.

PR #1262, opened 2026-06-29
Parse Claude Code extraction JSON arrays robustly

yangyihang96:
## Summary

Fixes Claude Code memory extraction when the backend LLM returns bracketed prose before the actual JSON array.

The parser currently slices from the first `[` to the last `]`. If the answer starts with harmless text like:

```text
Extraction notes [draft]: checked the transcript.
[{"type":"decision","title":"Queue","content":"Use a durable queue."}]
```

the parser starts at `[draft]`, fails JSON parsing, and drops the valid memories that follow.

## Reproduction

Before the fix, this response shape returns no memories because parsing starts at `[draft]` instead of scanning for the first valid JSON array:

```python
from claudecode_memanto.extractor import parse_llm_memories

text = """Extraction notes [draft]: checked the transcript.
[{"type":"decision","title":"Queue","content":"Use a durable queue."}]
"""

assert parse_llm_memories(text) == []
```

After the fix, the same input returns the valid `decision` memory from the second line. The added test covers that exact bracketed-prose failure mode.

## Fix

- Replace first-bracket/last-bracket slicing with `JSONDecoder.raw_decode` scanning.
- Return the first syntactically valid JSON array in the response.
- Add regression coverage for bracketed prose before the real array.

## Validation

```bash
uv run --with pytest --with-editable . --with-editable integrations/claudecode pytest integrations/claudecode/tests/test_extractor.py -q
uv run --with pytest --with-editable . --with-editable integrations/claudecode pytest integrations/claudecode/tests -q
uv run --with ruff --with-editable . --with-editable integrations/claudecode ruff check integrations/claudecode/claudecode_memanto integrations/claudecode/hooks integrations/claudecode/tests
git diff --check
```

Results:

- 15 focused extractor tests passed
- 69 Claude Code integration tests passed
- `All checks passed`
- `git diff --check` clean

Refs #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON array extraction by locating and decoding the first valid JSON array within the LLM response, even if bracketed text appears beforehand.
  * Reduced parsing failures by returning only the exact detected JSON array.
* **Tests**
  * Added a unit test to verify bracketed prose preceding the JSON array is ignored and that parsing still succeeds.

### DIFF
--- a/integrations/claudecode/claudecode_memanto/extractor.py
+++ b/integrations/claudecode/claudecode_memanto/extractor.py
@@ -148,11 +148,11 @@ def _extract_json_array(text: str) -> str | None:
         if char != "[":
             continue
         try:
-            parsed, end = decoder.raw_decode(cleaned[start:])
+            parsed, end = decoder.raw_decode(cleaned, start)
         except json.JSONDecodeError:
             continue
         if isinstance(parsed, list):
-            return cleaned[start : start + end]
+            return cleaned[start:end]
     return None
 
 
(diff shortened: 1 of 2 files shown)


Neelpatel1604:

PR #1269, opened 2026-06-29
fix: event-loop blocking in restart endpoint + zero memories_created (#770)

Kukrushka:
## Summary

Two bugs found via issue #770 bug & exploit challenge.

### Bug 1 — Event-loop freeze in `POST /api/ui/onprem/restart` (Architecture Flaw)

**File:** `memanto/app/ui/routes/ui_router.py` — `restart_onprem_backend()`

**Root cause:** Four blocking calls inside an `async def` freeze the entire FastAPI event loop for up to **6 minutes**, making the server completely unresponsive to all other requests during a backend restart:

| Line | Blocking call | Max wait |
|------|--------------|----------|
| `subprocess.run(["moorcheh", "down"], ...)` | synchronous subprocess | 60 s |
| `subprocess.run(up_args, ...)` | synchronous subprocess | 300 s |
| `_httpx.get(health_url, ...)` | sync HTTP inside async | 2 s × 60 |
| `time.sleep(1.0)` | sleep inside async | up to 60 s |

**Fix:**
- `subprocess.run` × 2 → `await asyncio.to_thread(subprocess.run, ...)`
- `httpx.get` → `async with httpx.AsyncClient() as http: await http.get(...)`
- `time.sleep(1.0)` → `await asyncio.sleep(1.0)`

**Reproduce:** Trigger `POST /api/ui/onprem/restart` and observe that concurrent API calls (memory writes, session ops) hang until the restart completes.

---

### Bug 2 — `memories_created` always returns `0` in `SessionSummary` (Logic Flaw)

**File:** `memanto/app/services/session_service.py:253`

**Root cause:** An unresolved `# TODO: Get actual memory count from backend` hardcodes `memories_created = 0`, so every call to `POST /agents/{agent_id}/deactivate` returns `"memories_created": 0` regardless of actual session activity.

**Fix:** In the route layer (`sessions.py`), snapshot the live Moorcheh namespace document count immediately before and after `end_session()`. The delta is assigned to `summary.memories_created`. The lookup is best-effort — if Moorcheh is unreachable the field stays `0` rather than raising.

**Reproduce:**
```bash
# Start session, write some memories, then deactivate
curl -X POST /api/v2/agents/{id}/activate ...
# write memories ...
curl -X POST /api/v2/agents/{id}/deactivate ...
# "memories_created" is always 0 in the response
```

## Files Changed

- `memanto/app/ui/routes/ui_router.py` — replace 4 blocking calls with async equivalents
- `memanto/app/routes/sessions.py` — compute real `memories_created` via namespace count delta

## Test Plan

- [ ] Restart on-prem backend; confirm other API endpoints remain responsive during the ~6 min wait
- [ ] Write N memories in a session, deactivate; confirm `memories_created == N` in the summary
- [ ] If Moorcheh is unreachable, confirm deactivate still succeeds with `memories_created == 0`

Closes #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Session deactivation now returns a more accurate newly created memories count by using live namespace item counts during session ending.
* **Improved Performance**
  * The on-prem restart flow no longer blocks the async event loop, and concurrent restart requests are serialized to avoid overlapping runs.
  * Health/readiness polling for the restart endpoint is now fully async, improving responsiveness and reliability.

### DIFF
--- a/memanto/app/routes/sessions.py
+++ b/memanto/app/routes/sessions.py
@@ -250,20 +250,10 @@ async def deactivate_agent(
         )
 
     try:
-        # Snapshot namespace document count before and after ending the session
-        # so end_session() receives an accurate memories_created value.
-        # _namespace_item_counts is best-effort; returns {} if Moorcheh is
-        # unreachable, so the count safely falls back to 0.
-        counts_before = await _namespace_item_counts(_server_api_key)
-        count_before = counts_before.get(session.namespace, 0)
-
-        counts_after = await _namespace_item_counts(_server_api_key)
-        count_after = counts_after.get(session.namespace, 0)
-        memories_created = max(0, count_after - count_before)
-
-        summary = get_session_service().end_session(
-            agent_id, memories_created=memories_created
-        )
+        # end_session_async() fetches the live Moorcheh namespace count and
+        # injects it as memories_created — logic owned by the service so all
+        # callers (HTTP + future CLI) get accurate counts.
+        summary = await get_session_service().end_session_async(agent_id)
         return summary
     except SessionNotFoundError as e:
         raise map_error_to_http_exception(e)
--- a/memanto/app/routes/sessions.py
+++ b/memanto/app/routes/sessions.py
@@ -250,7 +250,17 @@ async def deactivate_agent(
         )
 
     try:
+        # Snapshot memory count before and after to compute memories_created.
+        # _namespace_item_counts is best-effort; falls back to 0 if unreachable.
+        counts_before = await _namespace_item_counts(_server_api_key)
+        count_before = counts_before.get(session.namespace, 0)
+
         summary = get_session_service().end_session(agent_id)
+
+        counts_after = await _namespace_item_counts(_server_api_key)
+        count_after = counts_after.get(session.namespace, 0)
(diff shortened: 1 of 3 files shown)


Kukrushka:
## CodeRabbit issues addressed

**Fixed in latest commit** (`0dd9623`):

### Restart lock cancellation safety

The previous code released the lock on handler cancellation while `moorcheh down/up` was still running in the worker thread. Fixed using `asyncio.shield()` + explicit wait:

```python
inner = asyncio.ensure_future(_do_restart_onprem_backend(...))
try:
    return await asyncio.shield(inner)
except asyncio.CancelledError:
    try:
        await inner  # hold the lock until subprocess actually finishes
    except Exception:
        pass
    raise
```

### `memories_created` service contract (already addressed in previous commit)

This was moved into `SessionService.end_session(self, agent_id, memories_created: int = 0)` in commit `6f2ce30`. The route now computes the delta via `_namespace_item_counts` and passes it directly to the service — the value is no longer patched on the returned summary. Other callers that don't have access to the Moorcheh API pass `memories_created=0` (the default), which is the correct behaviour for non-HTTP paths (e.g. CLI) where the namespace item count is not measurable.

PR #1270, opened 2026-06-29
Restore memory when update reupload fails

drsteinerdj:
## Summary
- Preserve the original memory document before `update_memory()` deletes it.
- If the replacement upload fails after deletion, best-effort restore the original document to the same namespace.
- Preserve the current Moorcheh metadata shape during rollback: `agent_id`, `memory_type`, comma-separated `tags`, provenance, TTL, and timestamps.
- If the restore upload also fails, surface that explicitly instead of implying rollback succeeded.
- Add regression coverage for both failed-reupload paths, including restore namespace and metadata payload assertions.

## Why
`update_memory()` uses a delete-and-recreate flow. Before this change, a successful delete followed by a failed replacement upload could leave the original memory missing. The fix keeps the successful path unchanged and adds rollback only for the failed replacement upload path.

The rollback path now restores the original stored record without resetting the metadata fields that are still part of the current core memory model. It also handles older stored records that lack flat `agent_id` metadata by deriving the agent from the namespace.

The rollback failure case is also explicit now: `rollback also failed` is reported when the replacement upload fails and the original document cannot be restored either.

## Validation
```text
.\.venv\Scripts\python.exe -m pytest tests/test_unit.py -q
28 passed

.\.venv\Scripts\python.exe -m ruff check memanto/app/services/memory_write_service.py tests/test_unit.py
All checks passed!

.\.venv\Scripts\python.exe -m ruff format --check memanto/app/services/memory_write_service.py tests/test_unit.py
2 files already formatted

git diff --check
passed
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory updates to preserve existing timestamps, TTL, type, and tags when editing saved entries.
  * Added a safer recovery path if an updated memory can’t be saved, helping restore the previous version instead of leaving data inconsistent.
  * Made handling of stored date and tag formats more robust, reducing update failures from mixed data formats.

* **Tests**
  * Added regression coverage for successful rollback recovery and for cases where rollback restoration also fails.

### DIFF
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -306,6 +306,38 @@ class MemoryWriteService:
                 if metadata.get("expires_at"):
                     updated_memory.expires_at = metadata["expires_at"]
 
+            original_memory = MemoryRecord(
+                id=memory_id,
+                type=metadata.get("type", "fact"),
+                title=existing_memory_data.get("title", "Restored Memory"),
+                content=existing_memory_data.get("content", ""),
+                scope_type=metadata.get("scope_type", "agent"),
+                scope_id=metadata.get("scope_id", "unknown"),
+                actor_id=metadata.get("actor_id", "unknown"),
+                source=metadata.get("source", "system"),
+                source_ref=metadata.get("source_ref"),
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #1274, opened 2026-06-29
Guard direct memory sync cached exports

guo-gy:
## Summary
- Make `DirectClient.sync_memory_to_project` refresh the export before copying `MEMORY.md`, matching the SDK path and the CLI docs.
- This forces the normal agent session validation path before any cached export can be copied.
- Add a regression test proving a stale cached export cannot be synced for an agent without an active validated session.

## Reproduction
Before this change, if `~/.memanto/exports/victim_memory.md` existed locally, `sync_memory_to_project("victim", project_dir)` copied that cached file directly into `project_dir/MEMORY.md` without calling `_get_validated_session_for_agent("victim")`. A caller that knew another local `agent_id` could therefore sync stale private memory cache contents without holding that agent session.

## Fix
The direct client now calls `export_memory_md(agent_id=..., limit_per_type=...)` first. That path validates the requested agent session and refreshes the cache before the copy happens.

## Validation
- `uv run --group dev pytest tests/test_cli.py::TestMEMANTOCLI::test_direct_memory_sync_requires_session_even_with_cached_export -q`
- `uv run --group dev pytest tests/test_cli.py::TestMEMANTOCLI::test_memory_export tests/test_cli.py::TestMEMANTOCLI::test_memory_sync tests/test_cli.py::TestMEMANTOCLI::test_direct_memory_sync_requires_session_even_with_cached_export -q`
- `uv run --group dev pytest tests/test_cli.py -q`
- `uv run --group dev ruff check memanto/cli/client/direct_client.py tests/test_cli.py`
- `git diff --check`

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory sync behavior by refreshing the cached export before copying it into a project’s `MEMORY.md`.
  * Prevented syncing cached memory when the session scope is invalid, so `MEMORY.md` is not created.
  * When neither a cached export nor a fresh export file is available, sync now raises `FileNotFoundError` (instead of silently returning an empty result).
* **Tests**
  * Added CLI integration coverage for cached-export session validation and missing-fresh-export error handling.

### DIFF
--- a/memanto/cli/client/direct_client.py
+++ b/memanto/cli/client/direct_client.py
@@ -1539,27 +1539,40 @@ class DirectClient:
             Dict with ``output_path``, ``total_memories``, ``source``
             (``"cache"`` or ``"fresh"``).
         """
-        self.export_memory_md(agent_id=agent_id, limit_per_type=limit_per_type)
+        export_result = self.export_memory_md(
+            agent_id=agent_id, limit_per_type=limit_per_type
+        )
 
         cache_path = Path.home() / ".memanto" / "exports" / f"{agent_id}_memory.md"
         target_path = Path(project_dir) / "MEMORY.md"
         target_path.parent.mkdir(parents=True, exist_ok=True)
 
-        if cache_path.exists():
-            # Copy the freshly updated cache.
-            shutil.copy2(str(cache_path), str(target_path))
-            content = cache_path.read_text(encoding="utf-8")
-            mem_count = content.count("### ")
-            return {
-                "output_path": str(target_path.resolve()),
-                "total_memories": mem_count,
-                "source": "cache",
-            }
+        source_path = cache_path
+        source_label = "cache"
+        if not source_path.exists():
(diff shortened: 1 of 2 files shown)


guo-gy:
Fixed the sync success path so it now verifies a readable export source exists, copies it to the project, and confirms `MEMORY.md` exists before returning success. Added a regression test for the missing fresh-export case.

Validation:
- `uv run pytest tests/test_cli.py -q`
- `uv run ruff check memanto/cli/client/direct_client.py tests/test_cli.py`

guo-gy:
Additional validation after the fix:
- `uv run pytest -q`

Only live Moorcheh E2E tests were skipped because this environment has the placeholder API key, matching the existing test skip guard.

@coderabbitai review

Xenogents:
Thanks for the pr! We're closing this because Memanto is assumed to be run entirely locally and by a single user, so there is no security risk giving access to local files.

PR #1275, opened 2026-06-29
Fix multi-scope memory search import

2y999:
# Fix multi-scope memory search import

## Summary

- Fixes `MemoryReadService.search_multi_scope()` importing `ScopeType` from `memanto.memanto.app.constants`, a non-existent package path.
- Adds a regression test covering multi-scope namespace construction and the Moorcheh query call.

## Bug

Calling `search_multi_scope()` currently fails before it can query memory because the method tries to import:

```python
from memanto.memanto.app.constants import ScopeType
```

That path does not exist in the package. This breaks the cross-agent / multi-workspace retrieval path, which is directly relevant to complex workflows and multi-agent use cases.

## Validation

```powershell
.venv\Scripts\python.exe -m pytest tests/test_unit.py::TestMemoryReadServiceMultiScope -q
.venv\Scripts\python.exe -m pytest tests/test_unit.py::TestMemoryWriteServiceDelete tests/test_unit.py::TestForgetEndToEnd -q
.venv\Scripts\python.exe -m pytest tests/test_unit.py -q
.venv\Scripts\python.exe -m ruff check memanto\app\services\memory_read_service.py tests\test_unit.py
.venv\Scripts\python.exe -m ruff format --check memanto\app\services\memory_read_service.py tests\test_unit.py
.venv\Scripts\python.exe -m py_compile memanto\app\services\memory_read_service.py tests\test_unit.py
```

All listed checks pass locally.

## Bounty

For `moorcheh-ai/memanto#770`: this is a reproducible core-package logic bug affecting multi-scope memory retrieval. The PR includes the fix and a focused regression test.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected multi-scope search namespace handling so searches use the proper scope constants and build the expected namespace list.

- **Tests**
  - Added coverage for multi-scope memory search, verifying namespace construction, search parameters, and returned results.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -175,7 +175,7 @@ class MemoryReadService:
             namespaces = []
             from typing import cast
 
-            from memanto.memanto.app.constants import ScopeType
+            from memanto.app.constants import ScopeType
 
             for scope_def in scopes:
                 scope = create_memory_scope(
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -268,6 +268,59 @@ class TestMemoryWriteServiceDelete:
         assert MemoryWriteService(client).delete_memory("m1", "ns") is expected
 
 
+class TestMemoryReadServiceMultiScope:
+    """Multi-scope search is the cross-agent retrieval path. It should build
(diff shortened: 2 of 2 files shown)


Xenogents:

PR #1279, opened 2026-06-29
fix(memory): timeline amnesia in search_as_of - add superseded_at field (#770)

q404365631:
# Fix: Timeline Amnesia in `search_as_of` — introduce dedicated `superseded_at` field

> Submitted for [Bounty #770 — Bug & Exploit Challenge](https://github.com/moorcheh-ai/memanto/issues/770) ($100 USD)

## Summary of the flaw

`MemoryReadService.search_as_of` is a **point-in-time query** ("What was true at this point in time?"). To decide whether a memory was already superseded at the requested `as_of_date`, it inspects the memory's `updated_at` field.

The problem: `updated_at` is **shared by every mutating operation** on a `MemoryRecord` — `validate()`, `detect_contradiction()`, `update_memory()`, etc. — not just `mark_superseded()`. Any later call to those operations "refreshes" `updated_at` to a much later timestamp, so `search_as_of` cannot tell when the supersession actually happened.

**Net effect — timeline amnesia**: a memory that was superseded *before* `as_of_date` silently reappears in the result set, causing the agent to recall a stale, already-replaced preference as if it were still true at that point in time.

This directly violates three of the bounty's in-scope targets:
- *Retrieval Quality & Accuracy* — fails to recall the correct version
- *Losing track of when an event occurred (timeline amnesia)*
- *Deeply flawed contradiction handling* — supersession is the system's primary contradiction-resolution mechanism

### Root cause (3 sub-causes)

1. **`search_as_of` uses `updated_at` to time supersession** — but `updated_at` is mutated by `mark_superseded`, `validate`, `detect_contradiction`, and `update_memory` alike.
2. **No dedicated field records *when* supersession happened.** The legacy `MemorySupersedeResponse` model has `supersede_timestamp`, but the core `MemoryRecord` does not, and `_mark_memory_superseded` in `app/legacy/universal_services.py` is a no-op stub.
3. **`validate()` / `detect_contradiction()` don't guard against `status == "superseded"`** — they happily bump `updated_at` on an already-replaced memory, which is exactly what poisons `search_as_of`.

## Reproduction steps

A self-contained PoC is included at [`docs/bounty_reports/poc_timeline_amnesia.py`](docs/bounty_reports/poc_timeline_amnesia.py). It requires only the `memanto` package importable — **no Moorcheh API key, no network** (the Moorcheh client is mocked).

```bash
cd <memanto repo root>
python docs/bounty_reports/poc_timeline_amnesia.py
```

### Timeline used by the PoC

| Time | Event |
|------|-------|
| 2026-01-01 | Memory A created ("user likes coffee") |
| 2026-02-01 | Memory B created; A is superseded by B |
| 2026-03-01 | A is (mistakenly) `validate()`-d → `updated_at` refreshed to 2026-03-01 |
| Query | `search_as_of(as_of_date="2026-02-15")` |

### Expected vs actual (pre-fix)

| | Expected | Actual (buggy) |
|---|----------|----------------|
| Result IDs | `["mem-B"]` | `["mem-A", "mem-B"]` |
| A present? | No (already superseded) | **Yes (wrongly recalled)** |

`search_as_of` returns the *already-replaced* preference A as if it were still true at 2026-02-15 — even though A was superseded on 2026-02-01.

Run the PoC on the pre-fix code → exit code `1` (bug reproduced). Run it on this branch → exit code `0` (bug fixed).

A pytest-based regression test is also included at `tests/failing_tests/test_timeline_amnesia.py` (9 tests, all passing on this branch).

## Fix / architectural proposal

**Central idea**: introduce a dedicated `superseded_at` field that is set *only* by `mark_superseded()` and never touched by `validate()` / `detect_contradiction()` / `update_memory()`.

### `memanto/app/core.py`
1. New field on `MemoryRecord`: `superseded_at: datetime | None = None`
2. `mark_superseded()` now sets `superseded_at = now` alongside `updated_at`.
3. `validate()` and `detect_contradiction()` early-return when `self.status == "superseded"` — no state change, no `updated_at` bump.
4. `to_moorcheh_document()` emits `superseded_at` so the field round-trips through Moorcheh storage.

### `memanto/app/services/memory_read_service.py`
1. `search_as_of` now reads `superseded_at` first, falling back to `updated_at` only for legacy data:
   ```python
   supersede_ts = memory.get("superseded_at") or memory.get("updated_at")
   ```
2. `_format_memory_item` extracts `superseded_at` so downstream consumers can see it.

### `memanto/app/services/memory_write_service.py`
`update_memory()` now preserves `superseded_by`, `supersedes`, and `superseded_at` from the existing memory (unless explicitly overridden in `updates`). Previously the delete-and-recreate pattern dropped these fields, which would have made a superseded memory look active again after an edit.

## Backwards compatibility

- **Old stored memories** (no `superseded_at` field): `search_as_of` falls back to `updated_at`, so behavior is unchanged — these memories continue to work, they just don't get the new protection until they are re-superseded.
- **No schema migration required**: `superseded_at` is an optional field; existing Moorcheh documents are not touched.
- **No API change**: no endpoint signature is modified; `superseded_at` is surfaced in response payloads only when present.

## Files changed

| File | Change |
|------|--------|
| `memanto/app/core.py` | New `superseded_at` field; `mark_superseded` sets it; `validate` / `detect_contradiction` skip superseded memories; `to_moorcheh_document` emits it |
| `memanto/app/services/memory_read_service.py` | `search_as_of` prefers `superseded_at`; `_format_memory_item` extracts it |
| `memanto/app/services/memory_write_service.py` | `update_memory` preserves `superseded_at` / `superseded_by` / `supersedes` |
| `tests/failing_tests/test_timeline_amnesia.py` | Regression test (was failing pre-fix, passes post-fix) |
| `docs/bounty_reports/poc_timeline_amnesia.py` | Self-contained PoC |
| `docs/bounty_reports/timeline_amnesia_supersession.md` | Detailed bug report |

## Verification

```bash
# PoC (pre-fix: exit 1, post-fix: exit 0)
python docs/bounty_reports/poc_timeline_amnesia.py

# Regression tests (9 tests, all pass on this branch)
python -m pytest tests/failing_tests/test_timeline_amnesia.py -v
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where older superseded memories could reappear in time-based search results after later updates.
  * Improved handling of supersession history so previously replaced items are filtered more reliably.

* **New Features**
  * Added support for preserving and displaying supersession timestamps in memory records and search output.

* **Tests**
  * Added regression coverage for supersession behavior, legacy data fallback, and timestamp preservation.

### DIFF
--- a/docs/bounty_reports/poc_timeline_amnesia.py
+++ b/docs/bounty_reports/poc_timeline_amnesia.py
@@ -1,0 +1,277 @@
+"""
+独立 PoC：验证 search_as_of 的 Timeline Amnesia bug (issue #770)
+
+不依赖 pytest，可直接 `python docs/bounty_reports/poc_timeline_amnesia.py` 运行。
+
+bug 描述：
+    MemoryReadService.search_as_of 用 memory['updated_at'] 判断 supersession
+    发生的时间。但 updated_at 会被 validate() / detect_contradiction() /
+    update_memory() 等多种操作刷新，导致 supersession 时间被错误判定为更晚，
+    进而让 point-in-time 查询返回已经被 supersede 的记忆。
+
+正确行为：
+    as_of_date 之前已被 supersede 的记忆不应出现在 search_as_of 结果中。
+
+运行方式：
+    cd <memanto repo root>
+    python docs/bounty_reports/poc_timeline_amnesia.py
+"""
+
+from __future__ import annotations
+
+import sys
+from datetime import datetime, timedelta, timezone
+from unittest.mock import MagicMock
+
+# 确保 memanto 包可被导入
+sys.path.insert(0, ".")
+
+from memanto.app.core import MemoryRecord
+from memanto.app.services.memory_read_service import MemoryReadService
+
+
+def iso(dt: datetime) -> str:
+    if dt.tzinfo is None:
+        dt = dt.replace(tzinfo=timezone.utc)
+    return dt.astimezone(timezone.utc).isoformat().replace("+00:00", "Z")
+
+
+def make_memory(
+    memory_id: str,
+    created_at: datetime,
+    updated_at: datetime,
+    *,
+    superseded_by: str | None = None,
+    status: str = "active",
+    title: str = "memory",
+    content: str = "content",
+) -> dict:
+    """构造 _format_memory_item 格式的 memory dict。"""
+    return {
+        "id": memory_id,
+        "title": title,
+        "content": content,
+        "text": f"[FACT] {title}\n\n{content}",
+        "type": "fact",
+        "memory_type": "fact",
+        "confidence": 0.9,
+        "status": status,
+        "tags": [],
+        "created_at": iso(created_at),
+        "updated_at": iso(updated_at),
+        "expires_at": None,
+        "ttl_seconds": None,
+        "actor_id": "user-1",
+        "source": "user",
+        "source_ref": None,
+        "scope_type": "agent",
+        "scope_id": "agent-1",
+        "score": 0.95,
+        "provenance": "explicit_statement",
+        "validation_count": 0,
+        "contradiction_detected": False,
+        "superseded_by": superseded_by,
+    }
+
+
+def main() -> int:
+    print("=" * 70)
+    print("PoC: Timeline Amnesia in search_as_of (issue #770)")
+    print("=" * 70)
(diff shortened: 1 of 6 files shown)


het0814:

PR #1293, opened 2026-06-30
Skip malformed memory read items

guo-gy:
Refs #770

## Summary
- skip non-object items returned from core memory read paths before formatting them
- make `get_memory()` continue past malformed rows and return the first valid non-expired memory
- treat non-object `metadata` payloads as empty metadata so flat fields can still be used
- add regression coverage for `get_memory`, paginated document reads, similarity-search result formatting, and malformed metadata

## Reproduction
`MemoryReadService` assumed every document/search item and nested `metadata` value had object shape. If a backend/client response included a truncated scalar row such as `"truncated row"`, `_format_memory_item()` called `.get()` on that scalar and the whole read failed before valid memories were returned. If a row had `metadata` as a scalar, flat fallback fields like `memory_type` were never reached.

The added tests reproduce mixed responses containing malformed rows plus valid memory documents. Before the fix, `get_memory()`, `search_recent()`, and `search_memories()` raise `MemoryError` or drop the valid flat-field fallback path.

## Fix
Filter core read-path items to object-shaped rows before formatting them, have `get_memory()` scan for the first valid non-expired item, and normalize scalar `metadata` to `{}` before flat-field fallback. Summary chunks and existing type/tag/TTL filters continue to behave as before.

## Validation
- `uv run pytest tests/test_unit.py::TestMemoryReadServiceMalformedItems -q`
- `uv run pytest tests/test_unit.py -q`
- `uv run ruff check memanto/app/services/memory_read_service.py tests/test_unit.py`
- `uv run python -m py_compile memanto/app/services/memory_read_service.py tests/test_unit.py`
- `git diff --check`
- `uv run pytest -q`

Live E2E tests were skipped locally because `MOORCHEH_API_KEY` is not configured in this environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory read and search to safely skip malformed/non-standard upstream entries and return the first valid non-expired memory.
  * Search results now ignore non-dictionary items and handle multi-scope result formatting more defensively.
  * Memory formatting is more resilient when metadata is missing or not in the expected format.
* **Tests**
  * Added unit tests covering malformed “truncated row” payloads for memory retrieval and search, ensuring only valid results are returned.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -42,14 +42,17 @@ class MemoryReadService:
 
             items: list[Any] = cast(list[Any], result.get("items", []))
             if items and isinstance(items, list) and len(items) > 0:
-                memory = self._format_memory_item(items[0])
+                for item in items:
+                    if not isinstance(item, dict):
+                        continue
+                    memory = self._format_memory_item(item)
 
-                # Apply TTL enforcement
-                filtered = self._filter_expired_memories([memory])
-                if filtered:
-                    return filtered[0]
-                else:
-                    return None  # Memory has expired
+                    # Apply TTL enforcement
+                    filtered = self._filter_expired_memories([memory])
+                    if filtered:
+                        return filtered[0]
+
+                return None
 
             return None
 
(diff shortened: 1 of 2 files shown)


Xenogents:
Thanks for you work! We like to employ a fail fast approach for moorcheh so that we can detect why the memory may end up malformed.

PR #1294, opened 2026-06-30
Guard batch summary result rows

guo-gy:
Refs #770

## Summary
- guard DirectClient and SdkClient batch summary logging against malformed `results[]` rows
- guard the API extraction route's session-summary logging against the same malformed rows
- preserve successful batch write responses even when a per-item result row is not object-shaped
- add regression coverage for CLI and API paths while session summary logging is enabled

## Reproduction
When `batch_remember()` or conversation extraction succeeds and a session is active, the code logs each stored memory to the local session Markdown summary. That logging path extracts per-memory ids with `batch_results[i].get("id")`. If a backend/client response contains a malformed scalar row such as `"truncated row"`, the write result has already been produced, but summary logging raises `AttributeError` before the caller receives the successful batch response.

The added tests reproduce a successful batch result with `results: ["truncated row", {"id": "mem-2"}]` while session logging is active. Before the fix, DirectClient, SdkClient, and the API extraction route can crash in the summary logging path.

## Fix
Check that each per-item result row is a dict before reading `id`. Malformed rows are logged with `memory_id=None`, while valid rows still keep their ids and the original batch result is returned unchanged.

## Validation
- `uv run pytest tests/test_api.py::TestMEMANTOAPI::test_extract_memories_from_conversation_stores_batch tests/test_api.py::TestMEMANTOAPI::test_extract_memories_ignores_malformed_batch_result_rows tests/test_cli.py::TestMEMANTOCLI::test_batch_remember_direct_client_ignores_malformed_result_rows tests/test_cli.py::TestMEMANTOCLI::test_batch_remember_sdk_client_ignores_malformed_result_rows -q`
- `uv run pytest tests/test_api.py tests/test_cli.py -q`
- `uv run ruff check memanto/app/routes/memory.py memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py tests/test_api.py tests/test_cli.py`
- `uv run python -m py_compile memanto/app/routes/memory.py memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py tests/test_api.py tests/test_cli.py`
- `git diff --check`
- `uv run pytest -q`

Live E2E tests were skipped locally because `MOORCHEH_API_KEY` is not configured in this environment.

### DIFF
--- a/memanto/app/routes/memory.py
+++ b/memanto/app/routes/memory.py
@@ -476,8 +476,9 @@ async def extract_memories_from_conversation(
         session_service = get_session_service()
         for index, record in enumerate(memory_records):
             batch_results = result.get("results", [])
+            batch_item = batch_results[index] if index < len(batch_results) else None
             memory_id = (
-                batch_results[index].get("id") if index < len(batch_results) else None
+                batch_item.get("id") if isinstance(batch_item, dict) else None
             )
             await asyncio.to_thread(
                 session_service.log_memory_to_session_summary,
--- a/memanto/cli/client/direct_client.py
+++ b/memanto/cli/client/direct_client.py
@@ -747,7 +747,8 @@ class DirectClient:
             batch_results = result.get("results", [])
 
             for i, mem in enumerate(memory_records):
-                mem_id = batch_results[i].get("id") if i < len(batch_results) else None
+                batch_item = batch_results[i] if i < len(batch_results) else None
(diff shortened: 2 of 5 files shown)


Xenogents:
Thanks for your work! There are some changes I wanted to make so I made another commit to a different branch and put you as co-author. This PR will still be considered for the bounty.

PR #1296, opened 2026-06-30
fix auto-renewed session token handoff

bbbbzzzzcc-afk:
## Summary

Fixes a session lockout caused by the HTTP API's automatic renewal path. When a session neared expiry, the server persisted a replacement session and invalidated the request token, but never returned the replacement token to the client. The triggering request succeeded; every later request using the only token the client knew failed with 401.

Bounty submission for #770.

## Reproduction

1. Activate an agent and retain the returned session token.
2. Put the session within `SESSION_EXTEND_THRESHOLD_MINUTES`.
3. Send any authenticated memory request. It succeeds and silently replaces the persisted session.
4. Repeat the request with the original token. It fails with 401 because that session ID is no longer active.

The new API regression test reproduces the full sequence and proves that the returned replacement token can continue making requests while the old token is rejected.

## Fix

- return the replacement JWT in the `X-Session-Token` response header
- expose that header through CORS
- update the web UI for both JSON and multipart requests
- update the TypeScript SDK before handling success or error responses
- document the raw HTTP client contract

## Validation

- `pytest -q` (all non-live tests pass; 24 live E2E tests skipped without a real Moorcheh key)
- `ruff check .`
- `ruff format --check .`
- `mypy .`
- TypeScript: 11 tests pass, typecheck passes, build passes

No credentials or secrets are included in this change.

## Summary by CodeRabbit

* **New Features**
  * Added automatic session renewal support, returning replacement session tokens via the `X-Session-Token` response header.
  * Web UI and the TypeScript SDK now automatically capture and use renewed session tokens for subsequent requests, including uploads.

* **Bug Fixes**
  * Improved cross-origin behavior: exposed renewal headers/credentials are now restricted according to configured allowed origins, especially when wildcard origins are enabled.

* **Documentation**
  * Documented how automatic session renewal works for HTTP clients and browser clients across origins.

* **Tests**
  * Added contract and SDK tests covering renewed tokens on both successful and error responses.

PR #1297, opened 2026-06-30
Fix session auto-renewal token propagation

2y999:
## Summary

Fixes a session auto-renewal reliability bug: `get_current_session()` can create and persist a renewed JWT when the active session is near expiry, but API clients never receive that replacement token. They continue using the old `X-Session-Token` until it expires, so auto-renewal silently fails from the client perspective.

This PR:
- returns the renewed JWT in `X-Session-Token` and marks the response with `X-Session-Renewed: true`
- updates the Web UI fetch helper to adopt renewed tokens automatically
- adds an API regression test that forces a near-expiry session and verifies the replacement token reaches the client

## Validation

- `python -m pytest tests/test_api.py::TestMEMANTOAPI::test_auto_renewed_session_returns_replacement_token -q`
- `python -m pytest tests/test_api.py -q`
- `python -m ruff check memanto/app/routes/auth_deps.py tests/test_api.py`
- `python -m ruff format --check memanto/app/routes/auth_deps.py tests/test_api.py`

## Summary by CodeRabbit

* **New Features**
  * Session renewals now return updated session information to the client automatically.
  * The app can refresh an active session token in responses and keep the user’s session state current.

* **Bug Fixes**
  * Improved handling of near-expired sessions so requests can continue seamlessly with a renewed token.
  * The interface now reflects when a session has been renewed, reducing unexpected sign-outs.

### DIFF
--- a/memanto/app/routes/auth_deps.py
+++ b/memanto/app/routes/auth_deps.py
@@ -4,7 +4,7 @@ Authentication Dependencies for V2 API
 Shared authentication utilities to avoid circular imports.
 """
 
-from fastapi import Header, HTTPException
+from fastapi import Header, HTTPException, Response
 
 from memanto.app.models.session import Session
 from memanto.app.services.session_service import get_session_service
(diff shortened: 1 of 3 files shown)


2y999:
Closing this in favor of #1296, which predates this PR and covers the same auto-renewed token propagation issue with a broader patch set. I'll focus follow-up work on non-duplicate bugs.

PR #824, opened 2026-06-25
Enforce upload file size limit

n4p68vc6p9-hub:
## Summary
- add an explicit 5GB upload limit for `/upload-file`
- reject oversize files with HTTP 413 before calling the Moorcheh upload API
- stream uploads to the temporary file in chunks instead of reading the whole file into memory
- add API regression coverage for oversize uploads

## Root cause
The upload route documented a maximum file size but accepted uploads without checking the actual size. It also used `await file.read()` to load the entire file into memory before writing to disk, so an oversized request could consume process memory and still be forwarded to the backend.

## Validation
- `python -m pytest tests/test_api.py::TestMEMANTOAPI::test_upload_file_rejects_oversized_file -q`
- `python -m pytest tests/test_api.py::TestMEMANTOAPI::test_upload_file_with_session tests/test_api.py::TestMEMANTOAPI::test_upload_file_unsupported_extension tests/test_api.py::TestMEMANTOAPI::test_upload_file_rejects_oversized_file tests/test_api.py::TestMEMANTOAPI::test_upload_file_requires_session tests/test_api.py::TestCWE200ApiKeyLeak::test_traversal_filename_is_sanitized tests/test_api.py::TestCWE200ApiKeyLeak::test_absolute_path_filename_is_sanitized -q`
- `python -m pytest tests/test_api.py -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced a 5GB maximum upload limit with early rejection (HTTP 413).
  * Uploads are streamed while writing temporary data, aborting once the limit is exceeded.
  * Oversized uploads now return consistent, API-friendly error details.

* **New Features**
  * Added request body size limiting middleware to reject oversized HTTP requests before reaching upload handlers.

* **Tests**
  * Added API and middleware unit tests to verify HTTP 413 responses for oversized uploads and streamed bodies without a Content-Length header.

PR #1298, opened 2026-06-30
Stream uploaded files with enforced size limit

2y999:
## Summary

Fixes an upload resource-exhaustion bug: `/api/v2/agents/{agent_id}/upload-file` documents a 5GB maximum file size, but the implementation called `await file.read()` before enforcing any size limit. A large request body could be loaded fully into server memory before being rejected or sent to Moorcheh.

This PR:
- streams uploads to the temporary file in 1MB chunks
- enforces the documented 5GB maximum while streaming
- returns HTTP 413 before calling `documents.upload_file` when the limit is exceeded
- adds a regression test that monkeypatches the limit to 4 bytes and verifies oversized uploads are rejected before the Moorcheh client is invoked

## Validation

- `python -m pytest tests/test_api.py::TestMEMANTOAPI::test_upload_file_with_session tests/test_api.py::TestMEMANTOAPI::test_upload_file_rejects_oversized_body tests/test_api.py::TestMEMANTOAPI::test_upload_file_unsupported_extension -q`
- `python -m pytest tests/test_api.py -q`
- `python -m ruff check memanto/app/routes/memory.py tests/test_api.py`
- `python -m ruff format --check memanto/app/routes/memory.py tests/test_api.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Large file uploads are now streamed in chunks instead of loading the full file into memory.
  * Uploads that exceed the allowed size are now rejected with a clear “payload too large” response.
  * HTTP errors during upload are now returned consistently without being replaced by a generic error.

### DIFF
--- a/memanto/app/routes/memory.py
+++ b/memanto/app/routes/memory.py
@@ -47,6 +47,8 @@ from memanto.cli.config.manager import ConfigManager
 router = APIRouter()
 
 _config_manager = ConfigManager()
+MAX_UPLOAD_BYTES = 5 * 1024 * 1024 * 1024
+UPLOAD_CHUNK_BYTES = 1024 * 1024
 
 
 class RecallRequest(BaseModel):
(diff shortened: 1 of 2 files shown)


2y999:
Closing this in favor of #824, which predates this PR and covers the same upload resource-exhaustion issue with server-level request limiting plus streaming checks. I'll focus follow-up work on non-duplicate bugs.

PR #1309, opened 2026-07-01
Fix update memory delete response handling for on-prem backends

wanqinghd:
## Summary

This fixes an edit-path inconsistency between `delete_memory` and `update_memory`.

`delete_memory` already accepts both cloud delete responses (`actual_deletions`) and on-prem delete responses (`deleted_ids` / `status`). However, `update_memory` uses a delete-and-recreate flow and only checked `actual_deletions`, so an on-prem backend could successfully delete the old document but still make the edit fail before the replacement upload.

Changes:
- Extract the delete response handling into `MemoryWriteService._deletion_processed()`.
- Reuse that helper in both `delete_memory` and `update_memory`.
- Add a regression test for the on-prem `deleted_ids` response shape in `update_memory`.

Bounty issue: #770

## Reproduction

1. Mock an existing memory returned from `documents.get`.
2. Mock `documents.delete` to return the on-prem shape:
   ```python
   {"status": "success", "deleted_ids": ["mem-abc"]}
   ```
3. Call `MemoryWriteService.update_memory(...)`.
4. Before this patch, it raises `MemoryError: Failed to delete old version of memory mem-abc` because `actual_deletions` is absent.
5. After this patch, the update proceeds to upload the replacement document and returns `action == "updated"`.

## Tests

- `uv run pytest tests/test_unit.py::TestMemoryWriteServiceUpdate::test_update_memory_handles_onprem_delete_response_shape -q`
- `uv run pytest tests/test_unit.py::TestMemoryWriteServiceDelete -q`
- `uv run pytest tests/test_unit.py tests/test_cli.py tests/test_api.py -q`
- `uv run ruff check memanto/app/services/memory_write_service.py tests/test_unit.py`
- `uv run pytest -q`

Full test run passed locally. Live E2E tests were skipped because no real `MOORCHEH_API_KEY` is configured, matching the repo's skip condition.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory update/delete success detection to recognize additional backend delete response formats.
  * Fixed cases where certain delete responses were previously misinterpreted, impacting update flows that delete and recreate entries.
  * Added a unit test to verify successful handling of an on-prem delete response shape and ensure the delete occurs before upload.

### DIFF
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -23,6 +23,18 @@ class MemoryWriteService:
         self.client = moorcheh_client
         self._parser = MemoryParsingService()
 
+    @staticmethod
+    def _deletion_processed(result: dict[str, Any]) -> bool:
+        """Return whether a Moorcheh delete response represents a processed delete."""
+        raw = result.get("actual_deletions")
+        if isinstance(raw, int):
+            return raw > 0
+        for key in ("deleted_ids", "requested_ids"):
+            ids = result.get(key)
+            if isinstance(ids, list):
+                return len(ids) > 0
+        return str(result.get("status", "")).lower() in {"success", "ok"}
+
     def store_memory(
         self, memory: MemoryRecord, context: dict[str, Any] | None = None
     ) -> dict[str, Any]:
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #1324, opened 2026-07-01
Fix TypeScript SDK stale session retry

naufalfx805-source:
## Summary

Bounty submission for #770.

The TypeScript SDK keeps a cached session token and currently fails permanently once the server rejects that token with a session auth error. This can happen after a token expires or after the server-side session state is replaced. The next SDK call throws instead of recovering, even though the SDK already knows how to activate the agent again.

This PR:
- retries session-scoped JSON requests once after a 401/403
- refreshes the SDK session by clearing the cached token and re-running bootstrap/activate
- applies the same single retry to multipart uploads
- adds a regression test proving the first stale-token request uses the old token, the SDK re-activates, and the retry uses the new token

This is complementary to token handoff fixes that consume a replacement token from response headers; this fallback covers cases where the only signal available to the SDK is the auth failure response.

## Validation

- \\
pm run typecheck\\ passes in \\sdks/typescript\\
- \\
pm test\\ could not start on this machine because local Node is \\20.11.0\\; the installed Vitest dependency chain imports \\
ode:util.styleText\\, which is unavailable in that Node build.

### DIFF
--- a/sdks/typescript/src/index.ts
+++ b/sdks/typescript/src/index.ts
@@ -436,17 +436,26 @@ export class Memanto {
       await this.lifecycle.start();
     }
     const baseUrl = this.lifecycle.baseUrl;
-    const headers: Record<string, string> = {
-      "Content-Type": "application/json",
+    const send = () => {
+      const headers: Record<string, string> = {
+        "Content-Type": "application/json",
+      };
+      if (requireSession) {
+        headers["X-Session-Token"] = this.sessionToken ?? "";
+      }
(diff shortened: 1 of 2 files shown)


naufalfx805-source:
Rebased this PR on the latest upstream/main and resolved the TypeScript SDK test conflict. The PR is mergeable now.

Local validation:
- npm run typecheck passes
- npm test is still blocked on my local Node v20.11.0 because the installed Vitest/Rolldown dependency imports node:util.styleText, which is unavailable in that Node build.

het0814:

PR #1326, opened 2026-07-02
Fix TypeScript lifecycle stop after child exit

imgold2023-dot:
## Summary

Fixes a TypeScript SDK lifecycle edge case where `ServerLifecycle.stop()` can hang if the spawned `uvx memanto serve` process has already exited before `stop()` is called.

## Problem

`stop()` currently skips only missing or `child.killed` processes. A child process that exits on its own can have:

- `killed === false`
- a non-null `exitCode` or `signalCode`
- an already-emitted `exit` event

In that state, `stop()` registers a new `once("exit")` listener after the event has already happened, then waits forever. This can make `Memanto.close()` hang in cleanup paths after a server crash or early server exit.

## Fix

Return immediately when the child process has already exited by checking `exitCode` and `signalCode` before attaching a new `exit` listener.

## Test

Added a regression test that:

1. Spawns a Node child process that exits immediately.
2. Waits until the process has already emitted `exit`.
3. Injects it into a `ServerLifecycle` instance.
4. Calls `stop()` and asserts it resolves instead of timing out.

## Verification

From `sdks/typescript`:

```bash
npm test -- test/lifecycle.test.ts
npm run typecheck
npm test
npm run build
```

All commands pass locally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shutdown handling so stopping a server no longer tries to act on processes that have already exited or been signaled.
  * `stop()` now returns promptly in these cases, reducing the chance of unnecessary delays during cleanup.
* **Tests**
  * Added coverage to confirm shutdown completes quickly when the tracked process has already ended.

### DIFF
--- a/sdks/typescript/src/lifecycle.ts
+++ b/sdks/typescript/src/lifecycle.ts
@@ -96,7 +96,13 @@ export class ServerLifecycle {
     const child = this.process;
     this.process = null;
     this.url = null;
-    if (!child || child.killed) return;
+    if (
+      !child ||
+      child.killed ||
+      child.exitCode !== null ||
+      child.signalCode !== null
+    )
+      return;
 
     let exited = false;
     await new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


imgold2023-dot:
Refs #770.

het0814:

PR #1328, opened 2026-07-02
fix: resolve three core memory bugs in format, TTL, and update paths

wongwk0825:
## Summary

This PR fixes three bugs discovered during the Memanto Bug Challenge (#770) that affect memory integrity, retrieval accuracy, and API stability.

### 1. 🐛 Tag metadata leaks into displayed content during memory recall

**Severity**: Medium — affects retrieval quality and UX

**Root Cause**: `_format_memory_item` in `MemoryReadService` used `split("\n\n", 2)` which only split into at most 3 blocks. When a Moorcheh document had 3+ paragraphs, the trailing Tags metadata block could merge with preceding content, causing tag strings (e.g., `"market, finance"`) to appear in the displayed content returned to users.

**Fix**: Split on all double-newline boundaries and only strip a trailing `Tags: ` block that appears in the final position. Content paragraphs that genuinely start with `Tags: ` are now preserved.

**Reproduction**:
```python
text = "[FACT] Market Size\n\nParagraph 1\n\nParagraph 2\n\nTags: market, finance"
# Before fix: content includes "Tags: market, finance"
# After fix: Tags are properly stripped
```

### 2. 🐛 TTL-based memory expiration non-functional when set via constructor

**Severity**: Medium — affects memory lifecycle management

**Root Cause**: `MemoryRecord` constructor accepted `ttl_seconds` but never computed `expires_at` because `set_ttl()` was only called from external code, not from the constructor. The TTL enforcement in `_filter_expired_memories` checks `expires_at`, not `ttl_seconds`, so TTL-based expirations were silently broken.

**Fix**: Added `model_post_init` to compute `expires_at` when `ttl_seconds` is set during construction, matching the behavior of `set_ttl()`.

### 3. 🐛 Memory update crashes with server error when preserving TTL

**Severity**: High — crashes the update API

**Root Cause**: `update_memory` copied `expires_at` from existing metadata as a raw string, but `to_moorcheh_document()` called `.isoformat()` on it, crashing with `AttributeError: str object has no attribute isoformat`.

**Fix**:
- `update_memory` now parses string expiry into datetime before storing
- `to_moorcheh_document()` defensively handles both datetime and string values

## Testing

- All 43 existing tests pass
- 10 new tests added covering all three fixes
- Test coverage includes:
  - Tag stripping with 3+ paragraphs
  - Non-trailing Tags content preserved
  - Tags-only block after title
  - TTL constructor expiration computation
  - String expires_at handling in serialization
  - Datetime expires_at still works

## Files Changed
- `memanto/app/core.py` — `model_post_init` for TTL, defensive `to_moorcheh_document()`
- `memanto/app/services/memory_read_service.py` — Fix tag stripping in `_format_memory_item`
- `memanto/app/services/memory_write_service.py` — Parse string expiry in `update_memory`
- `tests/test_memory_core_fixes.py` — 10 new regression tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory expiration handling by supporting both string and datetime expiry values, including safer conversion when TTL metadata is present.
  * Made exported memory documents resilient when expiry is stored as text, ensuring consistent, non-crashing serialization.
  * Refined how memory entries are parsed into title and content, improving handling of multi-paragraph text and `Tags: ` blocks to avoid accidental stripping.
* **Tests**
  * Added focused test coverage for expiry serialization and title/content parsing edge cases.


Xenogents:
Thanks for your work! Your fixes will be taken into account and though we're preemptively merging this PR it will be considered for the bounty.

PR #1332, opened 2026-07-02
Filter secret-like conversation memory candidates

renaissancefieldlite:
Linked bounty: moorcheh-ai/memanto#770

## Summary
- add a local sensitive-value filter to conversation memory extraction normalization
- skip LLM-generated candidates that contain credential-like values before they can become stored memories
- add a regression test proving that a secret-like extracted candidate is dropped while a safe candidate from the same response is retained

## Why
The extraction prompt already tells the model not to include secrets, API keys, passwords, or tokens. The normalizer still accepted those candidates if the model returned them anyway. This creates a memory-integrity/security hardening gap: secret-like values can be persisted even though the prompt explicitly forbids them.

This patch does not rely on the LLM obeying the prompt. It adds a deterministic local guard for common sensitive-value shapes such as API-key assignments, `sk-`/`mch-` tokens, GitHub tokens, AWS access key IDs, JWTs, and private-key blocks.

## Reproduction / before-state
I added `test_extract_drops_secret_like_candidates` first and ran it against the unpatched code. It failed because this candidate was accepted:

```text
The production API key is sk-test-1234567890abcdef.
```

The value is fake and used only as a deterministic regression marker.

## Verification

```bash
.venv/bin/python -m pytest tests/test_conversation_memory_extraction.py tests/test_memory_parsing.py -q
# 19 passed

.venv/bin/python -m ruff check memanto/app/services/conversation_memory_extraction_service.py tests/test_conversation_memory_extraction.py
# All checks passed
```

## Boundary
- no live credentials used
- no attack against Moorcheh infrastructure
- no backend probing
- this is a local package hardening fix with reproducible tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory extraction now ignores entries that appear to contain secret-like values, such as API keys, tokens, credentials, or private keys.
  * Improved candidate filtering so sensitive text is less likely to be stored as a conversation memory.

* **Tests**
  * Added coverage to verify that secret-like candidates are dropped while normal memory candidates are still retained.

### DIFF
--- a/memanto/app/routes/memory.py
+++ b/memanto/app/routes/memory.py
@@ -430,6 +430,18 @@ async def extract_memories_from_conversation(
                 "count": len(candidates),
             }
 
+        if not candidates:
+            return {
+                "agent_id": agent_id,
+                "session_id": session.session_id,
+                "dry_run": False,
+                "candidates": [],
+                "total_submitted": 0,
+                "successful": 0,
+                "failed": 0,
+                "results": [],
+            }
+
         write_service = MemoryWriteService(client)
 
         from typing import cast
--- a/memanto/app/services/conversation_memory_extraction_service.py
+++ b/memanto/app/services/conversation_memory_extraction_service.py
@@ -23,7 +23,7 @@ class ConversationMemoryExtractionService:
     SENSITIVE_VALUE_PATTERNS: ClassVar[tuple[re.Pattern[str], ...]] = (
         re.compile(
             r"\b(?:api[_\s-]?key|secret|password|passwd|token|credential|authorization|bearer)\b"
-            r"\s*(?:[:=]|is|was)\s*\S+",
+            r"\s*[:=]\s*\S+",
             re.IGNORECASE,
         ),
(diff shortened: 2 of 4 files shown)


renaissancefieldlite:
Pushed a follow-up commit addressing the review notes:

- Tightened the generic secret key/value regex to explicit `:` / `=` delimiters so normal prose like token-budget notes is not dropped.
- Sensitive-only extraction now returns an empty successful result instead of surfacing as an error.
- The `remember/extract` route now treats an empty candidate set as a clean no-op and avoids batch-writing an empty list.

Validated locally with:

```bash
.venv/bin/python -m pytest tests/test_conversation_memory_extraction.py tests/test_memory_parsing.py tests/test_api.py::TestMEMANTOAPI::test_extract_memories_from_conversation_dry_run tests/test_api.py::TestMEMANTOAPI::test_extract_memories_from_conversation_stores_batch tests/test_api.py::TestMEMANTOAPI::test_extract_memories_from_conversation_sensitive_only_noops -q
.venv/bin/python -m ruff check memanto/app/services/conversation_memory_extraction_service.py memanto/app/routes/memory.py tests/test_conversation_memory_extraction.py tests/test_api.py
```

Neelpatel1604:

PR #1339, opened 2026-07-03
[codex] Stream Claude Code transcript parsing

duct-tape2:
## Summary

- Stream Claude Code hook transcripts line by line instead of materializing the whole JSONL file with `readlines()`.
- Preserve the existing behavior: find the first `/skill` across the full transcript while returning only the bounded tail for distillation.
- Add a regression test with a streaming-only file object so the reader does not depend on `readlines()`.

## Bounty

Submission for #770.

## Why

`read_transcript_for_distillation()` only needs the first skill mention plus the last `max_messages`/`max_chars`, but the previous implementation loaded the entire transcript into memory before parsing. Large Claude Code sessions can make the async Stop hook unnecessarily memory-heavy or cause it to skip distillation after a read failure, which hurts memory stability and context continuity.

## Reproduction

The regression test in `integrations/claudecode/tests/test_hooks.py` uses a line-iterable transcript object whose `readlines()` method raises. On the previous implementation, `_read_transcript_full()` caught that failure and returned `(None, "")`, so the Stop hook would silently lose the skill tag and transcript text instead of distilling the session. With this patch, the same stream is parsed line by line and returns the expected skill plus bounded tail text.

## Validation

- `uv run --group dev pytest integrations/claudecode/tests/test_hooks.py -q`
- `uv run --group dev pytest integrations/claudecode/tests -q`
- `uv run --group dev ruff check integrations/claudecode/hooks/_common.py integrations/claudecode/tests/test_hooks.py`
- `git diff --check`

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Transcript processing now handles large Claude Code logs more efficiently by reading them incrementally instead of loading everything at once.
  * Improved reliability when parsing transcripts from streaming or iterable file sources.
  * Better error handling ensures transcript rendering fails gracefully and logs useful debugging details when parsing issues occur.

### DIFF
--- a/integrations/claudecode/hooks/_common.py
+++ b/integrations/claudecode/hooks/_common.py
@@ -17,6 +17,7 @@ Input fields follow the official Claude Code hooks reference: common fields are
 from __future__ import annotations
 
 import json
+import logging
 import os
 import re
 import sys
@@ -28,6 +29,8 @@ from typing import Any
 # Make the sibling ``claudecode_memanto`` package importable whether or not the
 # example has been pip-installed.
 sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
+
+logger = logging.getLogger(__name__)
 
 
 def run(main: Callable[[], int]) -> None:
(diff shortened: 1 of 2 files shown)


duct-tape2:
Quick status update for reviewers: this PR is ready from my side. CodeRabbit re-reviewed the follow-up and reported no actionable comments, and the earlier logging suggestion was addressed in `092d137`.

The validation commands in the PR description cover the touched hook/test files. Happy to adjust the scope if you prefer a smaller or different fix for #770. Thanks!

PR #1352, opened 2026-07-05
fix: only strip Tags: suffix from content when metadata tags exist

Meet-hybrid:
Description
This PR fixes a bug where content containing a line starting with "Tags: " was being silently truncated during read-back in memory_read_service.py.

The Problem
The previous logic incorrectly assumed that any line starting with "Tags: " at the end of a memory item was an automatically appended metadata suffix. This caused legitimate user-generated content to be stripped whenever the flat tags field was empty.

The Fix
Updated the conditional logic in _format_memory_item to ensure the content-stripping behavior only triggers when the tags field is actually populated. This prevents legitimate text from being removed.

Testing & Verification

Added two new tests in tests/test_memory_format.py to verify that content is preserved when tags are empty and that existing metadata suffixes are still correctly handled.

Confirmed all 8/8 tests in test_memory_format.py pass, alongside full unit, integration, and API test suites.

Verified clean ruff linting.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory formatting so titles and content are extracted more reliably from multi-paragraph text.
  * Prevented paragraphs that begin with “Tags:” from being incorrectly removed when no tag metadata is present.
  * Preserved proper handling of entries with and without a type prefix.

* **Tests**
  * Added regression coverage for memory formatting to verify content integrity and tag parsing across several edge cases.

issue #770

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -657,30 +657,35 @@ class MemoryReadService:
         content = raw_text
 
         if raw_text:
-            lines = raw_text.split("\n\n", 2)  # Split into at most 3 parts
+            # Split into all \n\n-delimited segments so that the Tags: suffix
+            # can be reliably identified as the *last* segment, even when the
+            # content itself contains multiple paragraphs (multiple \n\n).
+            lines = raw_text.split("\n\n")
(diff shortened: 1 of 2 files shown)


Neelpatel1604:

PR #1357, opened 2026-07-05
[codex] Stream uploaded memory files with a size guard

ArisLiWind:
## Summary
- Stream uploaded memory files to temporary storage instead of loading the full file into memory.
- Enforce the documented 5GB upload limit while reading chunks and return a 413 response before calling the Moorcheh SDK.
- Preserve FastAPI HTTPException responses instead of remapping them to generic 500 errors.

## Why
The upload endpoint documented a 5GB maximum, but it used await file.read(), which loads the entire request body into memory before writing a temp file. Large uploads can exhaust memory or stall the server before the SDK sees the file.

## Validation
- HOME=/private/tmp/memanto-test-home .venv/bin/python -m pytest tests/test_api.py::TestMEMANTOAPI::test_upload_file_with_session tests/test_api.py::TestMEMANTOAPI::test_upload_file_rejects_oversized_file_before_sdk_upload

## Summary by CodeRabbit

* **Bug Fixes**
  * Large file uploads are now rejected cleanly with a clear 413 response when they exceed the maximum allowed size.
  * Uploads now stream in smaller chunks instead of loading the entire file into memory, improving reliability for big files.
  * Existing HTTP errors are now passed through unchanged, preserving the original error response.

* **Tests**
  * Added coverage for oversized upload handling to confirm invalid files are blocked before further processing.

### DIFF
--- a/memanto/app/routes/memory.py
+++ b/memanto/app/routes/memory.py
@@ -48,6 +48,9 @@ router = APIRouter()
 
 _config_manager = ConfigManager()
 
+MAX_UPLOAD_FILE_SIZE = 5 * 1024 * 1024 * 1024
+UPLOAD_CHUNK_SIZE = 1024 * 1024
+
 
 class RecallRequest(BaseModel):
     query: str = Field(..., min_length=1, description="Search query")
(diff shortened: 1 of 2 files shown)


Xenogents:
Thanks for your contributions! We're closing this PR as these fixes have already been addressed by other PR's.

PR #1360, opened 2026-07-05
Fix tag filters for comma-separated metadata

xiaobai0916:
## Summary
- normalize comma-separated tag metadata by trimming whitespace and dropping empty tags
- apply the same cleanup to list-based tag metadata
- add a regression test for `search_as_of(..., tags=["beta"])` matching metadata stored as `"alpha, beta"`

## Tests
- `python -m pytest tests/test_unit.py::TestMemoryReadServiceTagFilters::test_as_of_matches_comma_separated_tags_with_spaces -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory search and filtering to handle tags more reliably, including comma-separated tags with extra spaces.
  * Kept memory lookup, time-based filtering, and search behavior consistent while refining how results are displayed.

* **Tests**
  * Added coverage for tag filtering with spaced tag lists.
  * Updated existing unit test text and expectations to match the latest behavior.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -381,7 +381,7 @@ class MemoryReadService:
         seen_ids: set[str] = set()
         memories: list[dict[str, Any]] = []
         for item in items:
-            # Skip summary chunks — only return real memory documents
+            # Skip summary chunks ? only return real memory documents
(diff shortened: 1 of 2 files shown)


Neelpatel1604:

PR #1364, opened 2026-07-05
Handle on-prem delete result during memory updates

xiaobai0916:
## Summary
- reuse a single delete-result success helper for memory update and delete paths
- let update_memory accept on-prem delete responses such as `deleted_ids`/`status`, matching delete_memory behavior
- add regression coverage proving an on-prem delete response still allows the updated memory to be uploaded

## Why this matters
`update_memory()` uses a delete-and-recreate flow. On on-prem backends, `documents.delete()` can report success with `deleted_ids` instead of cloud-only `actual_deletions`. The previous update path treated that successful delete as a failure, so editing an existing memory could fail after the old version was deleted.

## Tests
- `python -m pytest tests/test_unit.py::TestMemoryWriteServiceDelete -q`
- `python -m ruff check memanto/app/services/memory_write_service.py tests/test_unit.py`
- `git diff --check`

Part of #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory update and delete handling so successful deletions are recognized across different response formats, including cloud and on-premise styles.
  * Fixed delete-and-recreate flows to proceed correctly when a deletion succeeds with alternative success indicators.

* **Tests**
  * Added coverage for successful update behavior when deletion responses return an on-premise-style success payload.

### DIFF
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -311,7 +311,7 @@ class MemoryWriteService:
                 namespace_name=namespace, ids=[memory_id]
             )
 
-            if delete_result.get("actual_deletions", 0) == 0:
+            if not self._delete_succeeded(delete_result):
                 raise MemoryError(f"Failed to delete old version of memory {memory_id}")
 
             validation_result = {"action": "store", "reason": "MVP direct store"}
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #1366, opened 2026-07-05
Preserve falsey metadata fields when formatting memories

xiaobai0916:
## Summary
- preserve valid falsey values from nested metadata when formatting memory items
- avoid falling back to flat fields for `confidence: 0.0`, empty status strings, or empty tag arrays
- add regression coverage for falsey metadata values

## Why this matters
`_format_memory_item()` preferred nested Moorcheh metadata, but used `metadata.get(field) or item.get(field)`. That means valid falsey metadata values were silently replaced by flat fallback values. A stored confidence of `0.0`, for example, could be displayed as a different flat value.

## Tests
- `python -m pytest tests/test_unit.py::TestMemoryReadServiceFormatting::test_format_memory_item_preserves_falsey_metadata_values -q`
- `git diff --check`

Part of #770.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -636,8 +636,11 @@ class MemoryReadService:
         def get_field(field_name, flat_field_name=None):
             """Get field from metadata object or fallback to flat field"""
             flat_name = flat_field_name or field_name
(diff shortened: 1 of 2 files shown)


Neelpatel1604:

PR #1368, opened 2026-07-06
Fetch extra search rows before TTL filtering

xiaobai0916:
## Summary
- fetch a buffered search window first, then retry with Moorcheh's page cap only when local filtering leaves the requested page underfilled
- prevent expired high-ranking rows from hiding valid lower-ranking memories
- add regression coverage for `limit=1` with expired rows ahead of the first valid result

## Bug / Impact
`search_memories()` sizes the backend request before applying local filters. TTL enforcement and temporal bounds are applied after retrieval. If the most similar rows are already expired and the caller asks for a small page such as `limit=1`, the service can request too few rows, filter them all out, and return no results even when a lower-ranked memory is still valid.

That hurts recall accuracy: a valid fact can be hidden by expired facts that happen to rank higher.

## Reproduction
The unit test reproduces the failure with eleven expired search rows followed by one valid row. The mock honors the requested `top_k`, so the old `limit + offset` sizing would only return expired rows and produce an empty response.

After the fix, `search_memories("answer", agent_id="agent-1", limit=1)`:
1. fetches the requested window plus a small buffer
2. applies temporal and TTL filtering locally
3. retries with `top_k=100` only if the filtered result set cannot fill the requested page
4. returns the valid memory after pagination

## Fix
Move the formatting and local filters into a small search helper, then use adaptive fetch sizing:
- initial request: `min(limit + offset + 10, 100)`
- retry request: `100`, only when the filtered results are still too short to satisfy `offset + limit`

This preserves the correctness fix while avoiding a full 100-row fetch for common queries that do not need it.

Related to #770.

## Tests
- `.\.venv\Scripts\python -m pytest tests\test_unit.py::TestMemoryReadServiceSearch::test_search_fetches_extra_rows_before_ttl_filtering -q`
- `.\.venv\Scripts\python -m pytest tests\test_unit.py -q`
- `.\.venv\Scripts\python -m ruff check memanto\app\services\memory_read_service.py tests\test_unit.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory search pagination so results stay accurate even when some items are filtered out by date or expiration rules.
  * Search pages now over-fetch when needed, helping ensure the requested number of valid results is returned.
  * Added coverage for expired-result handling to prevent pagination gaps.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -14,6 +14,8 @@ from memanto.app.config import settings
 from memanto.app.core import agent_namespace
 from memanto.app.utils.errors import MemoryError
 
+SEARCH_OVERFETCH_BUFFER = 10
+
 
 class MemoryReadService:
     def __init__(self, moorcheh_client: "MoorchehClient"):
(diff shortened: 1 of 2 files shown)


xiaobai0916:
Addressed the performance concern: search_memories now starts with the requested window plus a small buffer and only retries with top_k=100 when local TTL/temporal filtering cannot fill the requested page. I also strengthened the regression test so the mock truncates by requested top_k and proves the retry path is needed.

Validation:
- .\\.venv\\Scripts\\python -m pytest tests\\test_unit.py::TestMemoryReadServiceSearch::test_search_fetches_extra_rows_before_ttl_filtering -q
- .\\.venv\\Scripts\\python -m pytest tests\\test_unit.py -q
- .\\.venv\\Scripts\\python -m ruff check memanto\\app\\services\\memory_read_service.py tests\\test_unit.py

xiaobai0916:
Follow-up: extracted the over-fetch buffer into a named SEARCH_OVERFETCH_BUFFER constant for easier tuning while preserving the adaptive retry behavior.

Validation:
- .\\.venv\\Scripts\\python -m pytest tests\\test_unit.py::TestMemoryReadServiceSearch::test_search_fetches_extra_rows_before_ttl_filtering -q
- .\\.venv\\Scripts\\python -m pytest tests\\test_unit.py -q
- .\\.venv\\Scripts\\python -m ruff check memanto\\app\\services\\memory_read_service.py tests\\test_unit.py
- git diff --check

het0814:

PR #1370, opened 2026-07-06
Normalize MCP recall tags from metadata strings

nonggde:
## Summary

Fixes a Memanto MCP tool-shaping bug where recall results with comma-separated tag metadata (for example `"alpha, beta"`) were exposed to MCP clients as a character list via `list(raw.get("tags"))`. That can mislead agent-side filtering, display, and downstream reasoning over recalled memories.

## Reproduction

Before this change, `_to_memory_hit({"tags": "alpha, beta"}).tags` returned characters instead of tag names. This can happen when a backend or SDK surface returns the stored flat metadata value rather than the already-normalized list shape.

## Fix

- Add `_normalize_tags()` for MCP memory-hit shaping.
- Split comma-separated tag strings and trim whitespace.
- Keep list inputs clean and default missing tags to an empty list.
- Add regression tests for string, list, and missing tag inputs.

## Validation

- `.\.venv\Scripts\python -m pytest integrations\mcp\tests\test_tools.py -q` -> 3 passed
- `.\.venv\Scripts\python -m pytest integrations\mcp\tests -q` -> 28 passed
- `.\.venv\Scripts\python -m ruff check integrations\mcp\memanto_mcp\tools.py integrations\mcp\tests\test_tools.py` -> passed

Bounty context: #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tag handling so memory results now return tags consistently as a clean list.
  * Comma-separated tag values are split and trimmed correctly, list values are cleaned up, and missing tags now default to an empty list.
* **Tests**
  * Added coverage for tag normalization across string, list, and missing-value cases.

### DIFF
--- a/integrations/mcp/memanto_mcp/tools.py
+++ b/integrations/mcp/memanto_mcp/tools.py
@@ -903,7 +903,7 @@ def _to_memory_hit(raw: dict[str, Any]) -> MemoryHit:
         title=raw.get("title"),
         content=raw.get("content"),
         confidence=raw.get("confidence"),
-        tags=list(raw.get("tags") or []),
+        tags=_normalize_tags(raw.get("tags")),
         status=raw.get("status"),
         source=raw.get("source"),
         source_ref=raw.get("source_ref"),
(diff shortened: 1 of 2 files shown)


Neelpatel1604:

PR #1371, opened 2026-07-06
Count ok batch upload responses as successful

xiaobai0916:
Closes #770

## Summary

Fixes a batch upload accounting bug in `MemoryWriteService.batch_store_memories`.

On-prem/self-hosted Moorcheh responses are already treated as successful when deletion returns `{"status": "ok"}`, but batch uploads only counted `"queued"` and `"success"` as successful. If the backend returned `{"status": "ok"}` after accepting a batch upload, every pending result was updated to `"ok"` while the summary reported:

```text
successful: 0
failed: 0
total_submitted: N
```

That makes successful batch writes look like an empty/no-op result to API and CLI callers.

## Changes

- Adds `"ok"` to the accepted upload success statuses.
- Normalizes status comparison with `lower()` so equivalent response casing is counted correctly.
- Adds a regression test for batch upload responses returning `{"status": "ok"}`.

## Tests

- `.venv\Scripts\python.exe -m pytest tests\test_unit.py -q`
- `.venv\Scripts\python.exe -m pytest tests\test_api.py -q`
- `.venv\Scripts\python.exe -m ruff check memanto\app\services\memory_write_service.py tests\test_unit.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Batch memory uploads now treat additional statuses (including `ok`) as successful when tallying completed items.
  * Batch upload failure detection is now case-insensitive, improving consistency across varying backend response formats.
* **Tests**
  * Added unit tests covering batch uploads with `ok` responses and case-insensitive `FAILED` responses to verify correct `successful`, `failed`, and per-item status reporting.

### DIFF
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -204,7 +204,7 @@ class MemoryWriteService:
                 for r in results
                 if str(r["status"]).lower() in SUCCESSFUL_UPLOAD_STATUSES
             )
-            failed = sum(1 for r in results if r["status"] == "failed")
+            failed = sum(1 for r in results if str(r["status"]).lower() == "failed")
 
             return {
                 "total_submitted": len(memories),
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -13,6 +13,8 @@ from memanto.app.services.memory_parsing_service import MemoryParsingService
 from memanto.app.utils.errors import MemoryError
 from memanto.app.utils.ids import generate_memory_id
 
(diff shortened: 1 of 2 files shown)


xiaobai0916:
Addressed the review note by normalizing failed-status counting as well, and added a regression test for uppercase FAILED batch upload responses.\n\nVerified locally:\n- .venv\\Scripts\\python.exe -m pytest tests\\test_unit.py -q\n- .venv\\Scripts\\python.exe -m pytest tests\\test_api.py -q\n- .venv\\Scripts\\python.exe -m ruff check memanto\\app\\services\\memory_write_service.py tests\\test_unit.py

PR #1399, opened 2026-07-07
fix(langgraph): retry setup after activation failure

davidmosiah:
/claim #770

## Bug

`create_memanto_tools()` retries a tool call by running `_do_setup()` after an operation fails. `_do_setup()` currently swallows `activate_agent()` errors but still sets `_setup_done = True` unconditionally.

If activation fails transiently, later `memanto_remember`, `memanto_recall`, or `memanto_answer` calls skip setup entirely and keep failing instead of trying to activate the agent again.

## Fix

Only mark `_setup_done` after `activate_agent()` completes without raising. `create_agent()` failures remain tolerated because that path can legitimately fail when the agent already exists.

## Regression test

Added `test_memanto_remember_retries_setup_after_activation_failure`, which reproduces:

1. initial memory call fails because the session is inactive
2. first setup attempt has a transient activation failure
3. a later tool invocation retries setup instead of treating setup as complete
4. activation succeeds and the memory write succeeds

## Validation

- `PYTHONPATH=../.. uv run --with pytest --with pytest-asyncio --with pytest-timeout pytest tests/test_tools.py -q` -> 9 passed
- `PYTHONPATH=../.. uv run --with pytest --with pytest-asyncio --with pytest-timeout pytest tests -q` -> 33 passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry behavior for Memanto tools when agent activation fails.
  * Activation is now treated as incomplete unless activation succeeds, so subsequent memory actions reattempt setup.
  * Added a unit test covering recovery after the first activation failure and verifying successful completion on retry.


Xenogents:

PR #1400, opened 2026-07-07
fix(memory-write): same-ID overwrite prevents data loss (#1384, #770)

chenlinxi890-spec:
## Summary

Fixes the **Update Rollback Vulnerability** reported in issue #1384 (Bounty #770).

## Bug

The update_memory() method used a **delete-then-upload** pattern. If delete succeeded but upload failed (network crash, timeout), the memory was **permanently lost**.

## Fix

Changed to **same-ID overwrite** pattern — upload replacement with the same memory_id, no separate delete step:

1. Retrieve existing memory.
2. Apply updates.
3. Upload replacement with the same memory_id (Moorcheh documents are immutable by ID; uploading with the same ID replaces the old document atomically on the server side).
4. Validate upload status.
5. Do not perform a separate delete step.

This eliminates the race condition entirely — there is no window where the old version is deleted but the new one hasn't been stored yet.

## Changes

- 1 file changed: memanto/app/services/memory_write_service.py
- +19/-11 lines (minimal change)
- Preserves upstream's _deletion_succeeded() improvement

## Related

- Bounty: #770
- Bug report: #1384
- Related PR: #1320 (ambiguity guard fix)

PR #1390, opened 2026-07-07
Preserve memory provenance during edits

bobwoo0621-lab:
## Summary
Fixes a provenance attribution bug in memory edits. Updating a memory uses a delete-and-recreate flow, but the recreated `MemoryRecord` did not preserve the existing `provenance`, causing imported/inferred/observed/corrected memories to be rewritten as `explicit_statement`.

## Why it matters
Provenance is part of Memanto's trust and attribution model. Losing it during edits can make imported or inferred memories appear as directly stated facts, weakening contradiction handling, source attribution, and auditability.

## Reproduction steps
1. Store or load a memory with `provenance="imported"` and `source_ref`.
2. Edit only its content.
3. Inspect the re-uploaded Moorcheh document.
4. Before this fix, `provenance` becomes `explicit_statement`.

## What changed
- Added a regression test for provenance preservation during `MemoryWriteService.update_memory`.
- Preserved existing `provenance` when rebuilding the updated memory record.

## Test results
- `.\.venv\Scripts\python -m pytest tests/test_unit.py::TestMemoryWriteServiceDelete::test_update_memory_preserves_existing_provenance -q`
- `.\.venv\Scripts\python -m pytest tests/test_unit.py -q`
- `.\.venv\Scripts\python -m pytest -q`

Full suite passed locally. Live Moorcheh E2E tests were skipped because no real API key was configured.

Fixes/addresses #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory updates now preserve the original provenance information when editing an existing item, falling back to a sensible default if none is present.
  * Added coverage to verify that updating a memory keeps its existing provenance and source reference intact.

PR #1385, opened 2026-07-06
Preserve provenance metadata on memory update (#770)

2lll5:
## Summary
- Fixes a #770 memory integrity bug where `update_memory()` rebuilt stored memories without copying their existing `provenance` metadata.
- Prevents partial updates from silently downgrading `validated`, `corrected`, or `inferred` memories to the `MemoryRecord` default `explicit_statement`.
- Adds a regression test that reproduces the metadata loss and verifies the uploaded replacement document preserves provenance.

## Repro
1. Store/read a memory whose metadata has `provenance: validated`.
2. Call `MemoryWriteService.update_memory()` with a partial payload, e.g. only `content`.
3. Before this fix, the replacement document is uploaded with `provenance: explicit_statement` because the rebuilt `MemoryRecord` omits provenance.
4. After this fix, the existing provenance is retained unless the caller explicitly updates it.

## Tests
- `python3 -m pytest tests/test_unit.py::TestMemoryWriteService::test_update_memory_preserves_existing_provenance -q`
- `python3 -m pytest tests/test_unit.py -q`

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved memory provenance when updating existing entries, so partial edits no longer overwrite the original source information.
  * Ensured provenance handling defaults appropriately and that explicitly provided provenance in an update takes precedence over stored metadata.
* **Tests**
  * Added unit tests covering provenance behavior for partial updates and explicit provenance overrides.

PR #1406, opened 2026-07-08
fix(memory-write): same-ID update preserves data and metadata (#770)

evoesco:
## Summary

Bounty #770 submission for the high-severity update rollback/data-loss flaw reported in #1384.

`MemoryWriteService.update_memory()` previously deleted the existing document before uploading its replacement. A timeout, network error, backend failure, or rejected upload after the delete could permanently remove the memory.

This PR replaces that destructive sequence with a same-ID overwrite:

1. Retrieve the existing memory.
2. Rebuild the replacement with the original `memory_id`.
3. Preserve the existing type/provenance metadata during partial edits.
4. Upload the replacement under that same ID.
5. Validate the upload result.
6. Never issue a separate `documents.delete(...)` during update.

## Reproduction

The regression test models the realistic failure path directly:

```bash
python -m pytest   tests/test_unit.py::TestMemoryWriteServiceUpdate::test_update_memory_timeout_does_not_delete_old_memory   -q
```

The test configures a successful delete response and then makes the replacement upload raise `TimeoutError("network timeout")`.

- Against the vulnerable delete-then-upload implementation, the test fails because `documents.delete()` was already called once before the timeout.
- On this branch, the timeout surfaces as `MemoryError` and `documents.delete.assert_not_called()` passes, so the old memory is never deliberately removed.

Additional regressions prove that:

- queued same-ID uploads do not call delete;
- failed/unknown upload statuses raise a specific `MemoryError`;
- mixed-case successful statuses such as `OK` are accepted;
- partial updates preserve `memory_type`, `provenance`, `source_ref`, and tags.

## Why This Fix

A backup-and-restore or upload-then-delete flow still leaves extra failure states: rollback can fail, temporary documents can leak, and a follow-up delete can target the wrong same-ID replacement.

Same-ID overwrite removes the destructive delete step from the update path entirely. It is smaller, keeps the public API unchanged, and preserves the caller-visible memory ID.

## Changes

- Remove the separate delete operation from `update_memory()`.
- Keep the original `memory_id` on the replacement document.
- Accept `success`, `queued`, or `ok` upload statuses case-insensitively via the centralized status set.
- Preserve already-typed `MemoryError` messages.
- Preserve existing memory type and provenance during partial updates.
- Add focused regression coverage for queued, failed-status, timeout, no-delete, and metadata-preservation paths.

## Validation

```bash
python -m pytest -q
python -m ruff check .
python -m ruff format --check .
python -m mypy memanto
git diff --check
```

Results on the current PR head:

- 303 non-live tests passed.
- 24 live Moorcheh E2E tests were skipped because no real API key is configured locally.
- Ruff lint and formatting passed across the repository.
- MyPy passed across 75 source files.
- `git diff --check` passed.
- The latest actionable CodeRabbit finding was reproduced, fixed in `3817a4e`, and covered by a regression test.

Current upstream `main` (`47e91c6`) is merged into this branch. The only merge conflict was in the shared unit-test file; both the upstream batch tests and this PR's update regressions were retained. GitHub now reports the PR as mergeable.

## Scope and Related Work

- #809 and #1364 improve delete-response handling but retain the destructive delete path.
- #1270 adds best-effort rollback after deletion; this PR avoids entering that recovery state.
- #1400 contains the minimal same-ID overwrite idea; this PR adds failure-path and metadata/provenance regressions.
- #828, #1385, and #1390 demonstrate the metadata/provenance drift risk that this PR also covers.
- #1420 and #1421 use staging or upload/delete flows with more failure states and broader scope.

## Reviewer Focus

The core invariant is simple:

> `update_memory()` must not delete the existing memory before or after attempting the same-ID replacement upload.

A failed status or upload exception must leave the old record untouched by Memanto, while a successful partial update must preserve the original memory semantics unless the caller explicitly changes them.

## Related

- Bounty: #770
- Original bug report: #1384

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory updates now overwrite using the existing memory ID instead of deleting and recreating records.
  * Failed, timed-out, or non-success uploads no longer remove the prior memory, and report errors more clearly.
  * Upload success is accepted case-insensitively (e.g., “OK”).
  * Updates preserve key metadata, including memory type, provenance, source references, and tags.

* **Tests**
  * Expanded unit coverage for overwrite behavior, failure/timeout handling, case-insensitive success statuses, and metadata preservation.

\n\n## Social / Technical Write-up\n\nPublic engineering thread documenting the data-loss failure mode, same-ID overwrite invariant, and regression approach:\n\n- https://x.com/evoesco_studio/status/2076705341933252982\n

### DIFF
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -345,6 +345,8 @@ class MemoryWriteService:
                 "updated_fields": list(updates.keys()),
             }
 
+        except MemoryError:
+            raise
         except Exception as e:
             raise MemoryError(f"Failed to update memory: {e}")
 
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -307,9 +307,11 @@ class MemoryWriteService:
                     updated_memory.expires_at = metadata["expires_at"]
 
 
-            # Step 3: Upload new version FIRST (before deleting old)
-            # This prevents data loss if delete succeeds but upload fails
-            validation_result = {"action": "store", "reason": "MVP direct store"}
+
+            # Step 3: Upload new version (same-ID overwrite semantics)
+            # Moorcheh documents are immutable by ID — upload with the same memory_id
+            # replaces the old document atomically on the server side. No local delete needed.
+            validation_result = {'action': 'store', 'reason': 'MVP direct store'}
 
             from typing import cast
 
@@ -320,23 +322,15 @@ class MemoryWriteService:
                 namespace_name=namespace, documents=[document]
             )
 
-            if upload_result.get("status") not in ("success", "queued", "ok"):
+            status = upload_result.get('status', 'unknown')
+            if status not in ('success', 'queued', 'ok'):
                 raise MemoryError(
-                    f"Failed to upload updated memory {memory_id}: {upload_result.get('status')}"
+                    f'Failed to upload updated memory {memory_id}: {status}'
                 )
 
-            # Step 4: Now safely delete old version (new version already stored)
-            from typing import Any
-            delete_result = cast(
-                dict[str, Any],
-                self.client.documents.delete(namespace_name=namespace, ids=[memory_id]),
-            )
-
-            if not self._deletion_succeeded(delete_result):
-                # New version is already stored safely above; old version deletion
-                # failure is non-fatal — the memory update is still correct.
-                pass
-
+            # Upload succeeded (or queued for async processing). The old version
+            # is replaced on the server by this same-ID upload — no local delete needed.
+            # Returning early avoids a race where we'd delete the just-uploaded doc.
             return {
                 "id": memory_id,
                 "namespace": namespace,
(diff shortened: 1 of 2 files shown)


evoesco:
Bounty review update:

- Added an exact network-timeout regression in `564a4af`: against the vulnerable flow it fails because `documents.delete()` was already called; on this branch the timeout surfaces while delete remains uncalled.
- Revalidated this PR against current upstream `main` (`2044fdf`) in an isolated merge audit: no conflicts, the full non-live test suite passed, and Ruff passed on the touched files. The 24 live Moorcheh E2E tests were skipped because no real API key is configured locally.
- Refreshed the PR description with explicit reproduction steps, validation commands, and the focused comparison with rollback/staging alternatives.
- CodeRabbit is green with no actionable comments on the latest head.

The branch remains mergeable and maintainer edits are enabled.

evoesco:
@coderabbitai review

Xenogents:

evoesco:
Thanks for the update. Could you clarify which PR(s) account for the same-ID overwrite/data-loss and metadata/provenance coverage from #1406?

Current `main` still appears to use delete-then-upload in `MemoryWriteService.update_memory`, and the overlapping PRs #1364 and #1400 were also closed. I’d like to understand which implementation supersedes this work and whether #1406 will still be considered in the #770 bounty evaluation.

PR #1408, opened 2026-07-08
fix: 3 unique bugs - data loss, authz bypass, memory leak (#770)

heidevary:
## Bounty Challenge Submission for #770

This PR fixes 3 unique bugs not covered by any existing PR:

### Bug 1: Data Loss in `update_memory` (HIGH)

**File:** `memanto/app/services/memory_write_service.py`

**Problem:** `update_memory()` uses a delete-and-recreate pattern that deletes the old memory **before** uploading the new version (lines 309-330). If the upload fails (network error, service unavailable, invalid document), the old memory is already permanently deleted with no rollback mechanism.

**Fix:** Swap the order — upload the new version first, then delete the old one. If the upload fails, the original memory remains intact.

### Bug 2: Authorization Bypass via `scope_type` in Safe Deletion (HIGH)

**File:** `memanto/app/legacy/safe_deletion.py`

**Problem:** `validate_deletion_request()` only validates `scope_id` against `authenticated_scope_id` (line 116), but does **not** validate `scope_type`. Since `scope_type` is used to compute the namespace (line 186), an attacker authenticated for `scope_id="abc"` with `scope_type="agent"` could send a deletion request with `scope_type="session"` + `scope_id="abc"` to delete memories from a completely different namespace (`memanto_session_abc` instead of `memanto_agent_abc`).

**Fix:** Added `authenticated_scope_type` parameter to `validate_deletion_request()` and `validate_and_delete_memories()`. The parameter is optional (defaults to `None`) for backward compatibility — when provided, it validates that `scope_type` matches.

### Bug 3: Memory Leak + Thread Safety in Rate Limiter (MEDIUM)

**File:** `memanto/app/utils/rate_limiting.py`

**Problem:** 
1. `RateLimiter.requests` is a `defaultdict(deque)` that creates a new deque for each unique `operation:agent_id` key. When timestamps expire and are popped, the empty deque is **never removed** from the dict. With many unique agents, this causes unbounded memory growth.
2. The `check_rate_limit()` method has a TOCTOU race condition: checking `len(request_times) < limit.requests` and appending are not atomic. In a multi-threaded ASGI server, concurrent requests can both pass the check and exceed the rate limit.

**Fix:** 
1. After cleaning expired timestamps, remove empty deques from the dict: `del self.requests[key]`
2. Wrapped the check-and-append logic in a `threading.Lock` for thread safety

### Testing
All fixes are minimal and surgical — no behavior changes for the happy path. The scope_type validation is backward compatible (optional parameter).

/claim #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened deletion authorization by optionally verifying that the requested scope type matches the authenticated scope type, returning a 403 on mismatch.
  * Improved memory update reliability by switching to an upsert-style overwrite (same memory ID) and retrying uploads up to 3 times with exponential backoff for transient errors.
  * Enhanced rate limiting reliability under concurrent traffic with thread-safe per-key tracking, improved expired-entry cleanup, and removal of empty request buckets to reduce retention.

### DIFF
--- a/memanto/app/legacy/safe_deletion.py
+++ b/memanto/app/legacy/safe_deletion.py
@@ -108,7 +108,8 @@ class SafeDeletion:
 
     @staticmethod
     def validate_deletion_request(
-        scope_type: str, scope_id: str, ids: list[str], authenticated_scope_id: str
+        scope_type: str, scope_id: str, ids: list[str], authenticated_scope_id: str,
+        authenticated_scope_type: str | None = None,
     ):
         """Validate deletion request"""
 
@@ -119,6 +120,16 @@ class SafeDeletion:
                 detail={
                     "error": "scope_mismatch",
                     "message": f"Cannot delete from scope {scope_id} when authenticated as {authenticated_scope_id}",
+                },
+            )
+
+        # 1b. Require correct scope_type to prevent cross-namespace deletion
+        if authenticated_scope_type is not None and scope_type != authenticated_scope_type:
+            raise HTTPException(
+                status_code=403,
+                detail={
+                    "error": "scope_type_mismatch",
+                    "message": f"Cannot delete from scope_type '{scope_type}' when authenticated as '{authenticated_scope_type}'",
                 },
             )
 
@@ -242,6 +253,7 @@ def validate_and_delete_memories(
     actor_id: str,
     request_id: str,
     moorcheh_client,
+    authenticated_scope_type: str | None = None,
 ) -> dict[str, Any]:
     """Main function for safe memory deletion"""
 
@@ -251,6 +263,7 @@ def validate_and_delete_memories(
         scope_id=scope_id,
         ids=ids,
         authenticated_scope_id=authenticated_scope_id,
+        authenticated_scope_type=authenticated_scope_type,
     )
 
     # Perform deletion with audit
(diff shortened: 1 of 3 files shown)


heidevary:
## Self-Check Fixes

After reviewing my own PR, I found and fixed two issues:

### 1. `memory_write_service.py` - Reverted broken swap
**Problem:** My original fix swapped the delete/upload order (upload first, then delete). However, since `update_memory()` creates the new `MemoryRecord` with `id=memory_id` (same ID as the old version), and `to_moorcheh_document()` includes `"id": self.id`, uploading first and then deleting by `memory_id` would **delete the newly uploaded version**.

**Fix:** Reverted to the original delete-then-upload order (correct for Moorcheh's no-in-place-update architecture). Added 3x retry with exponential backoff for the upload step to mitigate the data loss risk if the upload fails after deletion.

### 2. `rate_limiting.py` - Fixed ineffective memory leak cleanup
**Problem:** My original fix did `del self.requests[key]` followed by `request_times = self.requests[key]`, but since `self.requests` is a `defaultdict(deque)`, the re-access immediately created a new empty deque, making the cleanup pointless.

**Fix:** Replaced with `.get(key)` to avoid defaultdict auto-creation. Empty deques are now properly deleted and the key is not re-accessed. New deques are only created when a request is actually allowed.

### 3. `safe_deletion.py` - Note on legacy code
The `authenticated_scope_type` validation fix is correct, but I want to be transparent: this module is in the `legacy/` directory and is not currently imported by any route handler. The actual deletion path uses `MemoryWriteService.delete_memory()` with session-based auth (`enforce_session_scope`). The fix is still valid for the legacy code and could be used if the module is re-activated.

PR #1414, opened 2026-07-09
fix(hermes): bind warmup to scheduled client

nexicturbo:
/claim #770

## Summary

Fixes a Hermes Memanto provider race where the background warmup thread could activate the wrong client after provider reinitialization.

Before this patch, `initialize()` scheduled `self._warmup`, and `_warmup()` dereferenced `self._client` when the thread actually ran. If Hermes reinitialized the provider for another session/profile before the first warmup thread executed, the first warmup could activate the newer session's client instead of the client it was created for.

This patch:

- binds the newly-created `_MemantoClient` to a local `client` variable during initialization
- passes that exact client into the warmup thread
- makes `_warmup()` activate the scheduled client instead of looking up mutable `self._client`
- adds a regression test that defers thread execution, initializes two sessions, then verifies the first scheduled warmup touches only the first client

## Verification

```text
New-Item -ItemType Directory -Force C:\tmp\codex-pytest-memanto | Out-Null; $env:TEMP = 'C:\tmp\codex-pytest-memanto'; $env:TMP = 'C:\tmp\codex-pytest-memanto'; $env:TMPDIR = 'C:\tmp\codex-pytest-memanto'; $env:PYTHONPATH = ".;integrations\hermes-agents"; .\.venv\Scripts\python.exe -m pytest integrations\hermes-agents\tests\test_provider.py -q
```

Result: `39 passed`

```text
.\.venv\Scripts\python.exe -m ruff check integrations\hermes-agents\hermes_memanto\provider.py integrations\hermes-agents\tests\test_provider.py
```

Result: `All checks passed!`

```text
.\.venv\Scripts\python.exe -m ruff format --check integrations\hermes-agents\hermes_memanto\provider.py integrations\hermes-agents\tests\test_provider.py
```

Result: `2 files already formatted`

```text
git diff --check -- integrations\hermes-agents\hermes_memanto\provider.py integrations\hermes-agents\tests\test_provider.py
```

Result: passed

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved background session warmup reliability so the correct connection is used even if the provider is reinitialized.
  * Reduced the chance of warmup affecting a newer session setup by binding it to the session created at startup.
* **Tests**
  * Added coverage for warmup behavior across repeated initialization.

### DIFF
--- a/integrations/hermes-agents/hermes_memanto/provider.py
+++ b/integrations/hermes-agents/hermes_memanto/provider.py
@@ -550,13 +550,14 @@ class MemantoMemoryProvider(MemoryProvider):
         if not self._api_key:
             return
         try:
-            self._client = _MemantoClient(
+            client = _MemantoClient(
                 self._api_key,
                 self._agent_id,
                 pattern=self._config["pattern"],
                 auto_create=self._config["auto_create"],
                 session_duration_hours=self._config["session_duration_hours"],
             )
+            self._client = client
             self._active = True
         except Exception:
             logger.warning("Memanto initialization failed", exc_info=True)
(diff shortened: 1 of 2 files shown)


nexicturbo:

PR #1415, opened 2026-07-09
fix(crewai): normalize recall tags for display

nexicturbo:
/claim #770

## Summary

Fixes a CrewAI recall-output formatting bug where backend-returned tag values that are not already `list[str]` can render incorrectly or raise.

Before this patch, `MemantoRecallTool._run()` did:

```python
tags = mem.get("tags", [])
tag_str = f" [tags: {', '.join(tags)}]" if tags else ""
```

If a backend/search result contains tags as a comma-separated string such as `"alpha,beta"`, the output becomes character-joined text (`a, l, p, h, a, ...`) instead of `alpha, beta`. If a scalar tag slips through, `join()` can fail.

This patch adds a small `_normalize_tags()` helper and uses it for:

- CrewAI remember tool tag parsing
- CrewAI recall output formatting for list/tuple/set, comma-separated string, scalar, or missing tag values

## Verification

```text
uv run --with pytest --with-editable . --with-editable integrations/crewai pytest integrations/crewai/tests/test_tools.py -q
```

Result: passed (`..........`)

```text
.\.venv\Scripts\python.exe -m ruff check integrations\crewai\crewai_memanto\tools.py integrations\crewai\tests\test_tools.py
```

Result: `All checks passed!`

```text
.\.venv\Scripts\python.exe -m ruff format --check integrations\crewai\crewai_memanto\tools.py integrations\crewai\tests\test_tools.py
```

Result: `2 files already formatted`

```text
git diff --check -- integrations\crewai\crewai_memanto\tools.py integrations\crewai\tests\test_tools.py
```

Result: passed

### DIFF
--- a/integrations/crewai/crewai_memanto/tools.py
+++ b/integrations/crewai/crewai_memanto/tools.py
@@ -37,6 +37,17 @@ VALID_MEMORY_TYPES = (
 )
 
 
+def _normalize_tags(raw: Any) -> list[str]:
+    """Normalize backend/user tag values for stable display and API calls."""
+    if raw is None:
+        return []
+    if isinstance(raw, str):
+        return [tag.strip() for tag in raw.split(",") if tag.strip()]
+    if isinstance(raw, (list, tuple, set)):
+        return [str(tag).strip() for tag in raw if str(tag).strip()]
+    return [str(raw).strip()]
+
+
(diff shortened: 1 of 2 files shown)


Neelpatel1604:

PR #1417, opened 2026-07-09
Avoid local storage writes during service initialization

WLLILLLL:
## Summary
This fixes a setup and CI friction issue where importing route modules or constructing local helper services can eagerly create files under the default data directory (~/.memanto). In restricted environments, read-only home directories, sandboxes, or import-only checks, that side effect can fail before the caller performs any state-changing operation.
## Problem
ConfigManager.__init__() created the config directory immediately.
AgentService.__init__() created the agents directory immediately.
SessionService.__init__() created the sessions directory immediately and generated/persisted a fallback JWT secret immediately when no secret was configured.
Route modules instantiate some of these helpers at import time, so the write can happen during collection/import instead of during an explicit save/create operation.
## Reproduction
Use an environment where the default home data path is not writable, or run tests/imports under a sandbox that denies writes to ~/.memanto.
Import modules that construct local helpers at module scope, or instantiate SessionService() without a configured secret.
The process can fail with a permission error while creating ~/.memanto, before any command actually saves config, agents, or sessions.
## Fix
Defer config directory creation until writing .env, config.yaml, or connections.json.
Defer agent directory creation until saving an agent; listing agents now returns an empty list when storage does not exist.
Defer session directory creation until saving session state or summary files.
Lazily generate the persisted fallback JWT secret only when token signing/verification needs a secret.
Add a regression test that constructing these services does not create local storage or a secret file.
Update existing tests that intentionally write session files to use explicit temporary storage.
## Validation
python -m pytest tests -q
Result:
Passing: all non-live tests.
Skipped: live Moorcheh E2E tests that require a real MOORCHEH_API_KEY.
Notes:
This is primarily a usability and setup-friction fix for import-time side effects. It should make the package more reliable in CI, sandboxed execution, read-only home environments, and applications that import Memanto modules before deciding whether to persist local state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when working with agents, sessions, and configuration by creating storage folders only when needed.
  * Prevents errors when listing agents or saving session data if the expected directories do not already exist.
  * Makes local setup less intrusive by avoiding automatic creation of files and folders during startup.

### DIFF
--- a/memanto/app/services/agent_service.py
+++ b/memanto/app/services/agent_service.py
@@ -29,7 +29,6 @@ class AgentService:
             agents_dir: Directory for agent metadata storage (defaults to ~/.memanto/agents/)
         """
         self.agents_dir = agents_dir or get_data_dir() / "agents"
-        self.agents_dir.mkdir(parents=True, exist_ok=True)
 
     def _generate_namespace(self, agent_id: str) -> str:
         """
@@ -134,6 +133,9 @@ class AgentService:
             AgentList with all agents
         """
         agents = []
+        if not self.agents_dir.exists():
+            return AgentList(agents=agents, count=0)
+
         for agent_file in self.agents_dir.glob("*.json"):
             with open(agent_file) as f:
                 data = json.load(f)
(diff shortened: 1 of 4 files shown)


het0814:

PR #1421, opened 2026-07-09
Restore conflict resolution and fix update_memory data loss (#1418)

### LINKED ISSUE #1418: Conflict resolution doesn't run — the "reconciles contradictions" feature is dead code
### Conflict resolution doesn't run — the "reconciles contradictions" feature is dead code
I went looking for how Memanto handles contradicting memories, since that's called out pretty heavily as a differentiator (both on memanto.ai and in the whitepaper, arXiv:2604.22085 Sec. E — the pitch is that when a new memory contradicts an old one, Memanto flags it and reconciles instead of just keeping both). Turns out it doesn't do that at all right now, on main / v0.2.4.

MemoryWriteService.store_memory() and batch_store_memories() never call any validation/conflict logic. It's just hardcoded:

```python

memanto/app/services/memory_write_service.py
skip validation for speed
Validate memory
validation_result = self.validation_service.validate_memory(memory, context)
Use validated memory if modified
if "memory" in validation_result:
memory = validation_result["memory"]
validation_result = {"action": "store", "reason": "MVP direct store"} ```

That comment made me assume there was a flag somewhere to turn it back on, but there isn't — the class it would call, ValidationPolicy, isn't defined anywhere in the repo anymore (grep -rn "class ValidationPolicy" comes up empty). The module that references it, memanto/app/legacy/memory_validation_service.py, throws an ImportError if you try to import it. So this isn't disabled, it's just gone — moved to legacy/ at some point and never reconnected. Same story for MemorySupersedeService in legacy/universal_services.py, which is presumably what would mark an old memory superseded — it's never called from the live write path either.

Practically: store "the deadline is April 15," then later "the deadline is April 22," and both sit there as active memories with unchanged confidence, forever. recall() has no way to know which one is current. This seems squarely in the "fails to resolve direct contradictions" / timeline-amnesia bucket from the bounty scope, and given it's a documented core feature that just silently doesn't run, I'd call it high severity rather than an edge case.

reproducing it
Doesn't need Docker or a Moorcheh key — the bug lives in plain Python logic, so the client is just mocked.

```python from unittest.mock import MagicMock from memanto.app.core import MemoryRecord from memanto.app.services.memory_write_service import MemoryWriteService

mock_client = MagicMock() mock_client.documents.upload.return_value = {"status": "success"} svc = MemoryWriteService(mock_client)

mem_v1 = MemoryRecord(type="preference", title="Favorite color", content="The user's favorite color is blue.", agent_id="agent-1", actor_id="user-1", source="user", confidence=0.9) mem_v2 = MemoryRecord(type="preference", title="Favorite color", content="The user's favorite color is red.", agent_id="agent-1", actor_id="user-1", source="user", confidence=0.9)

r1 = svc.store_memory(mem_v1) r2 = svc.store_memory(mem_v2)

both r1 and r2 come back memory_status == "active", confidence == 0.9, unchanged
from memanto.app.legacy.memory_validation_service import MemoryValidationService

ImportError: cannot import name 'ValidationPolicy' from 'memanto.app.core'
```

Full script + output attached.

a second thing I ran into while I was in there
Not directly related but found it tracing the same code path, so flagging it here rather than opening a separate issue: update_memory() (backs memanto edit) deletes the old document before it knows the new one uploaded successfully:

```python

Step 3: Delete old version
delete_result = self.client.documents.delete(...) ...

Step 4: Upload new version
upload_result = self.client.documents.upload(...) # if this throws, old one is already gone ```

If step 4 fails — timeout, 5xx, whatever — there's no rollback and no re-insert. The memory is just gone. Repro'd this with a mocked delete that succeeds and an upload that raises ConnectionError; update_memory re-raises but the delete already happened. Attached as a second script.

what I'd probably do about it
Not 100% sure this is the right shape without knowing more about the intended architecture, but roughly:

Bring back some version of ValidationPolicy (or replace it) and actually call it from the write path instead of leaving the call commented out. A reasonable starting point: run a similarity search scoped to the same memory_type before storing, and if something looks like a high-similarity conflict, mark the new one provisional or supersede the old one — there's already a _check_repetition stub in the legacy validation service that's most of the way there.
Hook MemorySupersedeService back up to that path instead of leaving it orphaned in legacy/.
For update_memory, flip the order — upload the new version first, only delete the old one once that's confirmed. If Moorcheh doesn't support that ordering for some reason (same ID collision, etc.) then at minimum a failed upload shouldn't leave you with nothing.
Would be good to have a regression test asserting a supersede/status change happens on conflicting stores, and that a failed update doesn't drop the original — both are easy to write against a mocked client, per the repros above.
Happy to take a stab at the actual PR if that's useful, just didn't want to guess at the intended design too much without a maintainer weighing in first.

Tested against main, installed via pip install -e ., Python 3.12 — this is all mocked-client behavior, doesn't touch the real backend either way.

[repro_conflict_resolution_bug.py](https://github.com/user-attachments/files/29840683/repro_conflict_resolution_bug.py)

[repro_update_data_loss.py](https://github.com/user-attachments/files/29840684/repro_update_data_loss.py)

CodeFrontGames:
Fixes #1418.

What was broken
Two issues in memory_write_service.py, both traced in #1418:

Conflict resolution didn't run at all. store_memory() and batch_store_memories() hard-coded {"action": "store", "reason": "MVP direct store"} and never called any validation or contradiction handling. The ValidationPolicy class that the validation service depended on had been removed from core.py, so app/legacy/memory_validation_service.py couldn't even be imported. Two directly contradictory memories were both stored active with unchanged confidence and no supersede link.

update_memory() could permanently lose data. It deleted the old document before confirming the new version uploaded. Any failure on the re-upload (timeout, backend 5xx) left the memory gone with no rollback.

What this changes
Re-adds ValidationPolicy to core.py — a lightweight, no-LLM policy keyed off settings.REQUIRE_VALIDATION_FOR. Critical memory types (fact, preference) that aren't corroborated or from a trusted source are stored provisional (capped confidence + TTL) rather than trusted outright.
Adds superseded_by to MemoryRecord.
Adds a working MemoryValidationService and a new ConflictResolutionService on the active services/ path. The orphaned legacy/ modules are left untouched.
Wires both into the write path: after a non-provisional memory is durably stored, any existing active, same-type, high-similarity memory (threshold via new CONFLICT_SIMILARITY_THRESHOLD setting, default 0.82) is marked superseded. Provisional (unconfirmed) writes deliberately don't supersede trusted data.
Rewrites update_memory() as stage-then-swap: upload the new content under a temporary id first, then delete the old doc, then promote to the real id, then clean up staging. A failure at any point leaves the content recoverable — never a state where the memory simply doesn't exist. The error message surfaces the staging id.
Design notes / open questions for maintainers
Conflict detection here is similarity-based, not semantic — it treats "another high-confidence, same-type memory about the same subject just arrived" as "the old one is stale," matching the freshness behavior already advertised, without adding an LLM call to every write. A stricter mode that actually asks a model "do these two conflict?" could be layered on top behind its own setting if you want higher precision at the cost of write latency. Happy to iterate on which behavior you'd prefer as the default.
Thresholds (CONFLICT_SIMILARITY_THRESHOLD, provisional caps) are exposed as settings so they're tunable without code changes.
Tests
New tests/test_conflict_resolution.py (5 tests): supersede-on-conflict, no-conflict-across-types, low-similarity-ignored, and two update_memory data-safety tests (failed promotion keeps data under staging id; successful update cleans up staging).
Updated the existing update_memory unit test to the new stage-then-swap contract.
Full suite passes locally (252 passed, 24 live-backend E2E skipped without an API key).
All reproducible against a mocked Moorcheh client — no live backend required.

## Summary by CodeRabbit

* **New Features**
  * Added similarity-based conflict detection for newly saved memories, with automatic superseding of older active matches.
  * Introduced conflict resolution configuration to tune the similarity threshold.
  * Added provisional storage for memory types that require additional validation before final persistence.
* **Bug Fixes**
  * Made memory updates safer via a stage-then-swap workflow to prevent data loss on partial failures.
  * Ensured update responses no longer include legacy validation details, returning update action/reason and updated fields instead.
* **Tests**
  * Added regression tests covering conflict superseding and staged-artifact recovery/cleanup on failed and successful updates.

### DIFF
--- a/memanto/app/config.py
+++ b/memanto/app/config.py
@@ -162,6 +162,11 @@ class Settings(BaseSettings):
     PROVISIONAL_TTL_SECONDS: int = 3600  # 1 hour
     PROVISIONAL_MAX_CONFIDENCE: float = 0.5
 
+    # Conflict Resolution Configuration
+    # Similarity score above which a new memory is treated as superseding
+    # an existing active memory of the same type in the same namespace.
+    CONFLICT_SIMILARITY_THRESHOLD: float = 0.82
+
     # Schedule Configuration
     MEMANTO_SCHEDULE_TIME: str = "23:55"
 
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -42,6 +42,9 @@ class MemoryRecord(BaseModel):
     # Provenance
     provenance: ProvenanceType = "explicit_statement"
 
+    # Set when this memory has been superseded by a newer, conflicting one.
+    superseded_by: str | None = None
+
     # Timestamps (auto-populated by server)
     created_at: datetime = Field(default_factory=datetime.utcnow)
     updated_at: datetime = Field(default_factory=datetime.utcnow)
@@ -89,6 +92,8 @@ class MemoryRecord(BaseModel):
             document["expires_at"] = self.expires_at.isoformat()
         if self.ttl_seconds:
             document["ttl_seconds"] = self.ttl_seconds
+        if self.superseded_by:
+            document["superseded_by"] = self.superseded_by
 
         return document
 
@@ -100,3 +105,50 @@ class MemoryRecord(BaseModel):
         """Set TTL and expiration"""
         self.ttl_seconds = seconds
         self.expires_at = datetime.utcnow() + timedelta(seconds=seconds)
+
+
+class ValidationPolicy:
+    """
+    Decides what happens to a new memory before it's persisted.
+
+    Memory types listed in settings.REQUIRE_VALIDATION_FOR ("fact",
+    "preference" by default) are treated as consequential enough that
+    they shouldn't be trusted at full confidence on a single, unconfirmed
+    mention. Everything else is stored as-is. This intentionally stays
+    simple (no LLM call) so it doesn't add latency/cost to every write.
+    """
+
+    def validate_memory(self, memory: "MemoryRecord", context: dict[str, Any]) -> dict[str, Any]:
+        from memanto.app.config import settings
+
+        if memory.type not in settings.REQUIRE_VALIDATION_FOR:
+            return {"action": "store", "reason": "Non-critical memory type; stored directly"}
+
+        repetition_count = context.get("repetition_count", 0)
+        trusted_source = memory.source in {"tool", "system"} and memory.confidence >= 0.85
+        corroborated = repetition_count >= 2 or memory.provenance in {"validated", "corrected"}
+
+        if trusted_source or corroborated:
+            return {
+                "action": "store",
+                "reason": (
+                    f"Critical memory type '{memory.type}' met validation "
+                    f"requirements (repetition={repetition_count}, source={memory.source})"
+                ),
+            }
+
+        return {
+            "action": "store_provisional",
+            "reason": (
+                f"Critical memory type '{memory.type}' has not been corroborated "
+                "(no repeated mention, not from a trusted source) - stored provisionally"
+            ),
+        }
+
+    def make_provisional(self, memory: "MemoryRecord") -> "MemoryRecord":
+        from memanto.app.config import settings
+
+        memory.status = "provisional"
+        memory.confidence = min(memory.confidence, settings.PROVISIONAL_MAX_CONFIDENCE)
+        memory.set_ttl(settings.PROVISIONAL_TTL_SECONDS)
+        return memory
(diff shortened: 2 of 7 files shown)


Xenogents:
Thanks for your contributions! We're closing the PR as the bugs you've addressed have already been covered by other PR's.

PR #1424, opened 2026-07-10
fix: memory filter injection, TOCTOU data loss in update_memory, TTL enforcement

biteqaq-maker:
## Summary

Three security/stability fixes for the Memanto Bug Challenge (#770).

### Fix 1: Filter Injection via Query String
memory_read_service.py:_build_filtered_query()
Concatenates user query with #key:value filter syntax without escaping. Malicious query like `foo #memory_type:secret` can inject arbitrary filters.
**Fix:** Escape # in user query before appending system filters.

### Fix 2: TOCTOU Data Loss in update_memory()
memory_write_service.py:update_memory()
Delete-then-create pattern causes permanent data loss if process crashes between delete and re-upload.
**Fix:** Create-before-delete: upload new version first, only delete old on success.

### Fix 3: TTL Enforcement Fail-Open
memory_read_service.py:_filter_expired_memories()
Fail-open on parse errors keeps expired memories in results.

Closes #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory search handling so query text cannot be mistaken for additional metadata filters.
  * Updated memory edits to create the new version before removing the previous one, helping preserve data during updates.
  * Memory update responses now provide clearer update details, including the fields that changed.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -452,10 +452,15 @@ class MemoryReadService:
             for key, value in metadata_filters.items():
                 filter_parts.append(f"#{key}:{value}")
 
-        # Combine query with filters
+        # Combine query with filters.
+        # CRITICAL: Escape '#' in the user query so attackers can't inject
+        # Moorcheh's #key:value filter syntax via malicious query strings.
+        # Without this, a query like "foo #memory_type:secret" would add a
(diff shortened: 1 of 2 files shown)


Xenogents:
Thanks for your contributions! We're closing this PR as these fixes have already been accounted for by other PR's

PR #1440, opened 2026-07-11
Fix two silent memory-integrity bugs in MemoryReadService (content wipe on "Tags:" + temporal filter fail-open)

learn-dumboo24:
Addresses the Memanto Bug & Exploit Challenge (#770).

Two reproducible, high-severity data-integrity bugs on the memory read path. Both are silent (no error surfaced), both fixed with a small diff plus regression tests. Repro is pure Python, no backend or API key.

### Bug 1: retrieved content silently wiped/corrupted (`_format_memory_item`)
`MemoryRecord.to_moorcheh_document` serializes as `"[TYPE] {title}\n\n{content}"` with an optional trailing `"\n\nTags: ..."`. The reader did `split("\n\n", 2)` and dropped any part starting with `"Tags: "`. Result:
- content that begins with `Tags: ` gives retrieved content `""` (total loss)
- multi-paragraph content leaks the `Tags:` line back into the body

Fix: take the title off the first blank line, then strip only a genuine trailing tags block, and only when the record actually has tags (exactly when the serializer appends it).

### Bug 2: one bad timestamp disables the whole time window (`_apply_temporal_filter`)
Each record's `created_at` was parsed inside a list comprehension wrapped in a single `try/except`. One missing or unparseable `created_at` raised mid-comprehension, so the entire `created_after`/`created_before` filter was skipped and out-of-window memories leaked (e.g. "last 7 days" returns years-old records). The comment `# Skip invalid timestamps` shows per-record was the intent.

Fix: parse each record defensively and skip only the offending record.

### Reproduction
```
pytest tests/test_memory_read_parsing_and_temporal.py
```
Fails on `main` (content=='' ; `['old','bad','june']` leak through), passes with this PR.

### Tests
7 regression tests, existing suite unaffected.

### Additional observations (not fixed here, flagged for maintainers)
- `cli/config/manager.py` `_normalize_duplicated_api_key` halves a valid even-length key whose two halves match, causing silent credential corruption.
- `app/services/session_service.py` `delete_session` lacks the `validate_safe_id` guard every sibling method has (latent path traversal, not reachable via current routes).

Happy to send follow-up PRs for either.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved date-range filtering so records with missing or invalid timestamps are excluded without disabling the requested filter.
  * Preserved content containing “Tags:” paragraphs while correctly extracting trailing metadata tags.
* **Tests**
  * Added regression coverage for timestamp filtering and content/tag parsing edge cases.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -476,31 +476,43 @@ class MemoryReadService:
         """
         from memanto.app.utils.temporal_helpers import parse_iso_timestamp
 
+        def _created_dt(record: dict[str, Any]) -> datetime | None:
+            """Parse a record's created_at, returning None if it is missing/invalid."""
+            raw = record.get("created_at")
+            if not raw:
+                return None
+            try:
+                return parse_iso_timestamp(str(raw))
+            except (ValueError, AttributeError, TypeError):
+                return None
+
         filtered = results
 
         if created_after:
             try:
                 after_dt = parse_iso_timestamp(created_after)
+            except (ValueError, AttributeError, TypeError):
+                after_dt = None
+            if after_dt is not None:
+                # A single record with a missing/unparseable timestamp is skipped
+                # individually; it must never disable the whole window filter.
                 filtered = [
                     r
                     for r in filtered
-                    if r.get("created_at")
-                    and parse_iso_timestamp(r["created_at"]) >= after_dt
+                    if (dt := _created_dt(r)) is not None and dt >= after_dt
                 ]
-            except (ValueError, AttributeError):
-                pass  # Skip invalid timestamps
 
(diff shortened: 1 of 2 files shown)


learn-dumboo24:
Thanks for the review. Added `test_embedded_tags_paragraph_with_real_tags` (commit 458411c) covering the case where content's first paragraph starts with `Tags: ` while the record also has a genuine trailing tags block. That is exactly why the parser uses `rpartition` on the last `\n\n` rather than matching any `Tags: ` line. All 7 tests pass.

Xenogents:
Thanks for your work! Your fixes will be taken into account and though we're preemptively merging this PR it will be considered for the bounty.

PR #1445, opened 2026-07-12
fix: report additional semantic search pages

phanmhung:
`search_memories()` requested exactly `offset + limit` rows, so it could never see the next result and always returned `has_more: false`.

This fetches one extra row, uses it to set `has_more`, and still returns only the requested page.

Added a regression test. Full test suite, Ruff, and mypy pass.

Related to #770.


Xenogents:

PR #1449, opened 2026-07-12
fix(client): prevent silent data overwrites in export_memory_md during backend outages

### LINKED ISSUE #1340: on-prem/cloud: export_memory_md silently overwrites MEMORY.md with empty data during a backend outage
**MEMANTO version:** 0.2.3
**Python version:** 3.14.6 (repro also applies to any 3.10–3.12 install per `pyproject.toml`)
**Operating system:** macOS (Darwin)

## Steps to reproduce

1. `memanto agent activate <agent-id>` with an agent that already has memories.
2. Make the backend briefly unreachable (on-prem: stop the local server; cloud: any transient network failure works too since the bug is in the client, not the transport).
3. Run `memanto memory sync` (or `memanto memory export`).

## Actual behavior

Every per-type `recall` call inside `export_memory_md` fails, but the exception is swallowed per type:

https://github.com/moorcheh-ai/memanto/blob/main/memanto/cli/client/sdk_client.py#L1344-L1354
(same pattern in `direct_client.py#L1485-L1495`)

```python
for mem_type in MEMORY_TYPE_ORDER:
    try:
        result = self.recall(...)
        memories_by_type[mem_type] = result.get("memories", [])
    except Exception:
        memories_by_type[mem_type] = []
```

With every type now `[]`, the function still writes the export unconditionally. `sync_memory_to_project` in `sdk_client.py` calls `export_memory_md` on *every* sync before copying the cache — so a single transient outage overwrites both the cached export (`~/.memanto/exports/{agent}_memory.md`) and the project's `MEMORY.md` with empty content. The CLI reports this as if it were a normal empty result:

```
No memories found for this agent.
Empty memory.md written to: ...
```

This happened to me mid-session: an unrelated local infra hiccup took the on-prem server down for a few seconds, and the very next `memanto memory sync` (fired by an editor hook) silently wiped a project's `MEMORY.md` that had real accumulated context — no error, no warning, just an "empty" result that looked like normal first-run output.

`direct_client.py`'s `sync_memory_to_project` is not hit every call (it only exports when no cache exists yet), but it has the same failure mode on the *first* sync for a new agent: if that first attempt races a backend hiccup, the empty result becomes the permanent cache, since nothing ever re-checks staleness afterward.

## Expected behavior

A total recall failure (every memory type errored) should be treated as "backend unreachable," not "zero memories," and should not silently overwrite good data. A partial failure (e.g. one type errors while others succeed) is fine to keep exporting, since that's very likely a genuine empty type rather than an outage.

## Fix

Opened a PR: generalizes on "if every type failed, raise `ConnectionError` instead of writing," and has `sync_memory_to_project` fall back to the previous export (`source: "stale-cache"`) when a refresh fails but a prior good export exists, instead of propagating the wipe.

santozion17-afk:
Fixes #1340.

### Problem
In both `SdkClient.export_memory_md()` and `DirectClient.export_memory_md()`, the exporter loops through all memory types to retrieve their values. If a transient backend outage or connection error occurs, the `self.recall()` call throws an exception for every type. 

Previously, these exceptions were swallowed inside the loop, causing the exporter to return empty memory lists for all types. This empty payload was then written to disk, silently wiping out existing cache files (`~/.memanto/exports/{agent}_memory.md`) and project `MEMORY.md` backups during an outage.

### Solution
We now track exceptions raised inside the recall loop. If *all* recall requests raise exceptions, we raise a `ConnectionError` instead of writing the output. This propagates the failure to the caller (e.g. `sync_memory_to_project`) and prevents silent data overwrites.

This has been successfully verified locally against the reproduction case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory exports now report a connection error when retrieving all memory types fails.
  * Exports continue with successfully retrieved memory types when only some requests fail.
  * Partial failures are no longer silently ignored.

### DIFF
--- a/memanto/cli/client/direct_client.py
+++ b/memanto/cli/client/direct_client.py
@@ -1503,6 +1503,7 @@ class DirectClient:
 
         memories_by_type: dict[str, list] = {}
 
+        exceptions = []
         for mem_type in MEMORY_TYPE_ORDER:
             try:
                 result = self.recall(
@@ -1512,8 +1513,15 @@ class DirectClient:
                     type=[mem_type],
                 )
                 memories_by_type[mem_type] = result.get("memories", [])
-            except Exception:
+            except Exception as e:
+                exceptions.append(e)
                 memories_by_type[mem_type] = []
+
+        if len(exceptions) == len(MEMORY_TYPE_ORDER):
+            raise ConnectionError(
+                f"Failed to export memories: Backend unreachable or all recall requests failed. "
+                f"First error: {exceptions[0]}"
+            )
 
         export_svc = self._get_export_service()
         out = output_path if output_path else None
--- a/memanto/cli/client/sdk_client.py
+++ b/memanto/cli/client/sdk_client.py
@@ -1365,6 +1365,7 @@ class SdkClient:
 
         memories_by_type: dict[str, list] = {}
 
+        exceptions = []
         for mem_type in MEMORY_TYPE_ORDER:
             try:
                 result = self.recall(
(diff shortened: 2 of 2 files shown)


Xenogents:
Thanks for your contributions! We're closing this PR as some of our recent updates have addressed this issue.

PR #1450, opened 2026-07-12
fix: Enforce strict SourceType validation and fail-closed rate limit

### LINKED ISSUE #1438: [BOUNTY] 🐜 Rate limiter fail-open + 2 validation inconsistencies
## Summary

During a security review of Memanto for the Bug & Exploit Challenge (#770), I identified three issues:

---

## 🔴 Finding 1: Rate Limiter Fail-Open Design (Severity: Medium)

**File:** `memanto/app/utils/rate_limiting.py:52-53`

```python
def check_rate_limit(self, operation: str, agent_id: str):
    if operation not in self.limits:
        return True, None  # FAIL-OPEN: unknown ops always pass
```

When `enforce_rate_limit` is called with an unrecognized operation, the request silently passes. `enforce_namespace_rate_limit(operation, agent_id)` builds keys as `f"namespace_{operation}"` — calling it with `operation="list"` creates `"namespace_list"` which is NOT in the limits dict, bypassing rate limiting entirely.

**Fix:** Fail-closed — raise ValueError or deny by default.

---

## 🟡 Finding 2: Inconsistent is_valid_memory_id

Two diverging implementations in `ids.py` and `safe_deletion.py`:
- `ids.py`: requires `_` in ID, length > 4
- `safe_deletion.py`: regex `^[a-zA-Z0-9_-]+$`, length >= 4

A memory ID like `abc-123` passes deletion validation but fails the general check.

---

## 🟡 Finding 3: SourceType = str (no validation)

`constants.py:21`: `SourceType = str` — unlike MemoryType/ProvenanceType which use `Literal`, SourceType accepts any arbitrary string. The source field in requests becomes searchable metadata with no validation.

---

## PoC

```python
from memanto.app.utils.rate_limiting import rate_limiter
allowed, _ = rate_limiter.check_rate_limit("nonexistent_op", "test")
assert allowed == True  # Bug confirmed
```

All reproducible against HEAD. Happy to submit PRs for any finding.

adamsithr:
Fixes #1438. Also fixed the tests that were using incorrect SourceTypes. 

Wallet (Solana): 7jXAK7dDBwkAJT67WjPeu8cg577y3x1oMp7ZCSMV18mR

## Summary by CodeRabbit

- **Bug Fixes**
  - Memory candidates are now consistently labeled with the `"system"` source.
  - Memory ID validation during deletion now uses the shared validation rules for more consistent results.
  - Unrecognized rate-limit operations now return a clear error instead of being silently allowed.

- **Improvements**
  - Source values are now restricted to supported categories, improving data consistency and preventing invalid classifications.

### DIFF
--- a/memanto/app/constants.py
+++ b/memanto/app/constants.py
@@ -18,7 +18,7 @@ MemoryType = Literal[
 ]
 
 # Source Types
-SourceType = str  # e.g., "user", "agent", "tool", "system", or specific "agent_name"
+SourceType = Literal["user", "agent", "tool", "system"]  # e.g., "user", "agent", "tool", "system", or specific "agent_name"
 
 # Status Types
 StatusType = Literal["active", "superseded", "deleted", "provisional"]
--- a/memanto/app/legacy/safe_deletion.py
+++ b/memanto/app/legacy/safe_deletion.py
@@ -144,8 +144,10 @@ class SafeDeletion:
                 },
             )
 
+        from memanto.app.utils.ids import is_valid_memory_id
+        
         # 4. Validate ID format
-        invalid_ids = [id for id in ids if not SafeDeletion._is_valid_memory_id(id)]
+        invalid_ids = [id for id in ids if not is_valid_memory_id(id)]
         if invalid_ids:
             raise HTTPException(
                 status_code=400,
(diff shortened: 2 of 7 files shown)


Xenogents:
Thanks for your changes! Your work will be considered when we're choosing a winner for the bounty.

PR #1348, opened 2026-07-03
fix: preserve extra metadata fields (e.g. original_id) on memory update

renaudgenard:
## Description

When `update_memory()` creates a new `MemoryRecord`, metadata fields that are not part of the `MemoryRecord` Pydantic schema (such as `original_id` in on-prem `data_store.json`) were silently dropped on upload.

This fix preserves any extra metadata fields from the existing record when building the updated document.

## Changes

- In `memory_write_service.py`, after converting the updated record to a Moorcheh document, copy over any metadata keys from the existing record that aren't already in the new document.

## Testing

- `to_moorcheh_document()` output unchanged for known MemoryRecord fields
- Extra fields (e.g. `original_id`) are now preserved across updates
- Backward compatible: no schema changes

Refs #1335

## Summary by CodeRabbit

* **New Features**
  * Added a new benchmark CLI with JSON or Markdown output for comparing memory frameworks.
  * Added a new benchmark dataset and documentation with quick-start instructions for offline and live runs.
* **Bug Fixes**
  * Preserved extra saved metadata during memory updates, helping records keep more information after edits.
* **Tests**
  * Added automated coverage for scoring, dataset loading, benchmark results, and CLI output.

PR #1454, opened 2026-07-13
fix: preserve original_id in update_memory() to prevent link breakage (on-prem)

### LINKED ISSUE #1335: Bug: update_memory() overwrites entire record including original_id — data_store.json not properly updated (on-prem)
## Bug Description (on-prem)

When calling `update_memory()` with an existing `memory_id` (from API `POST /items/batch`), the entire record is overwritten including `original_id`. This breaks the expected logic — if a record was originally created with one ID and later updated, the original ID should be preserved.

## Expected Behavior

- `update_memory()` should only update mutable fields: `data`, `tags`, `description`, `score`, `updated_at`
- `original_id` and `created_at` should remain unchanged
- `data_store.json` should update the specific record by `memory_id` without overwriting the entire structure

## Actual Behavior

The record in `data_store.json` is completely overwritten, including the original ID. On subsequent updates, `original_id` becomes equal to the latest `memory_id`.

## Steps to Reproduce

1. Create a record via API `POST /items/batch` with `memory_id=uuid1`
2. Backup `data_store.json`
3. Update the record via `POST /items/batch` with new `memory_id=uuid2`
4. Compare `data_store.json` — `original_id` has changed to `uuid2`

## Impact

- Loss of metadata on updates
- Incorrect conflict scanner logic
- Difficulties with audit trail (change history)

## Proposed Fix

In `memory_write_service.py`, the `update_memory()` method should:
1. Load the existing record from `data_store.json`
2. Update only the mutable fields
3. Preserve `original_id` and `created_at` unchanged

## Additional Notes

In the current implementation (SDK 0.2.4), there is already a partial fix — the delete-and-recreate block was removed from `update_memory()`, but the `original_id` overwrite problem remains.

## Configuration (on-prem)
- Memanto SDK: 0.2.4
- Python: 3.12
- OS: Ubuntu 24.04
- Deployment: systemd + direct binary (`memanto serve --port 7007`)
- Conflict Scanner: Moorcheh (Docker, port 8080)

---
Russian issue: https://github.com/moorcheh-ai/memanto/issues/1334

ardha27:
## Problem

`update_memory()` uses delete-and-recreate pattern but drops `original_id` from the existing record metadata when building the new `MemoryRecord`. This breaks downstream consumers that rely on `original_id` for linking related memory chains.

## Fix

Preserve `original_id` from the existing record metadata before constructing the new `MemoryRecord`. When `original_id` exists in the existing data, it is copied into the metadata dict used for field defaults, ensuring it survives the delete-and-recreate cycle.

## Related

Closes #1335

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved the original memory identifier when updating existing memories, ensuring continuity across delete-and-recreate operations.

### DIFF
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -285,6 +285,12 @@ class MemoryWriteService:
                 )
 
             # Build updated memory record
+            # Preserve original_id if it exists (prevents breaking links on update)
+            preserved_original_id = existing_memory_data.get("original_id")
+            metadata_for_construction = dict(metadata)
+            if preserved_original_id:
+                metadata_for_construction["original_id"] = preserved_original_id
+
             updated_memory = MemoryRecord(
                 id=memory_id,  # Keep same ID
                 type=updates.get("type", metadata.get("type", "fact")),


ardha27:
Closing as a duplicate of #1348, whose fix is already included in the maintainer aggregation #1428 with regression coverage and filtering for removed trust fields. This PR current metadata_for_construction value is not passed into MemoryRecord, so it does not add effective behavior beyond the existing fix.

PR #1462, opened 2026-07-14
fix(hermes): isolate Unicode agent identities

LC-C-hub:
## Summary

Fixes cross-profile memory namespace collisions in the Hermes integration. Fully non-ASCII profile identities previously all sanitized to `hermes`, which could make unrelated profiles share Memanto memory.

The fix NFKC-normalizes IDs and adds a stable SHA-256 suffix whenever non-ASCII identity data is removed, while preserving existing ASCII-only IDs and the 64-character limit. Regression tests cover Chinese, Cyrillic, mixed-script, and canonically equivalent Unicode identities.

## Validation

- Hermes provider tests: 41 passed
- Core suite: passed (24 live E2E tests skipped without an API key)
- Ruff: passed
- Mypy: passed
- `git diff --check`: passed

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Hermes-to-Memanto agent ID sanitization with Unicode NFKC normalization.
  * Reduced collisions by generating deterministic, bounded hashed suffixes for non-ASCII and mixed-script identities.
  * Ensured canonically equivalent Unicode inputs produce the same sanitized agent ID.
* **Tests**
  * Added regression coverage for non-Latin isolation, mixed-script hashing, malformed/surrogate handling, Unicode equivalence, and correct separation across provider instances.

### DIFF
--- a/integrations/crewai/crewai_memanto/tools.py
+++ b/integrations/crewai/crewai_memanto/tools.py
@@ -15,7 +15,7 @@ from crewai.tools import BaseTool
 from pydantic import BaseModel, Field
 
 from memanto.app.utils.errors import AgentAlreadyExistsError
-from memanto.cli.client.sdk_client import SdkClient
+from memanto.cli.client.sdk_client import SdkClient, sanitize_agent_id
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


Xenogents:
Thanks for you work! We're rejecting this PR as user wise we think having the potential for namespace collisions is better than adding random hashes to the end of namespace names.

PR #1469, opened 2026-07-15
fix: coalesce concurrent TypeScript server starts

ceigt:
/claim #770

## Summary

Coalesce concurrent `ServerLifecycle.start()` calls so the TypeScript SDK starts only one local Memanto server process.

## Flaw and impact

`start()` checked only the resolved `url`. During first startup, it awaits a free port and the health endpoint before setting that field. Two callers entering during that window therefore each selected a port and spawned a separate `uvx` process. The second spawn overwrote `this.process`, leaving the first process untracked and impossible for `stop()` to terminate.

This can happen when multiple public SDK methods are invoked concurrently on a fresh client, causing leaked local server processes and unnecessary resource usage.

## Reproduction

1. Create one `ServerLifecycle` without `baseUrl`.
2. Call `await Promise.all([life.start(), life.start()])` before either startup resolves.
3. Observe that `spawn()` is called twice on the previous implementation.

The new regression test performs this sequence with a mocked healthy child process and asserts a single spawn.

## Fix

- Cache the in-flight startup promise and return it to concurrent callers.
- Clear the cached promise after a startup failure so a later call can retry.
- Clear it during `stop()` so the lifecycle can start again normally.

## Validation

- `npm test -- --run test/lifecycle.test.ts` — 5 passed
- `npm run typecheck` — passed
- `npm test` — 32 passed
- `npm run build` — passed (ESM, CJS, and declarations)

Refs #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server startup reliability by coalescing concurrent start requests into a single in-flight startup.
  * Cleared stale startup state after failed attempts to ensure reliable retries.
  * Improved shutdown behavior by coordinating stop with any ongoing startup and avoiding waits on invalid spawn states.
* **Tests**
  * Expanded lifecycle coverage using deterministic mocked process spawning, including concurrency, timeout, and failure-mode scenarios.

### DIFF
--- a/sdks/typescript/src/lifecycle.ts
+++ b/sdks/typescript/src/lifecycle.ts
@@ -122,7 +122,9 @@ export class ServerLifecycle {
     if (this.starting === starting) {
       this.starting = null;
     }
-    if (!child || child.killed) return;
+    if (!child || child.killed || child.pid === undefined || child.exitCode !== null) {
+      return;
+    }
 
     let exited = false;
     await new Promise<void>((resolve, reject) => {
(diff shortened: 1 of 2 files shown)


het0814:

PR #1475, opened 2026-07-15
fix: preserve agent when remote purge fails

dicnunz:
## Summary

DELETE /api/v2/agents/{agent_id}?delete-backup-too=true currently catches every remote namespace deletion error, removes local agent/session state anyway, and returns a success message claiming all memories were deleted.

This changes the operation to fail before local deletion when the requested remote purge fails. The local agent remains available, so the user can retry instead of losing the only normal handle to the retained namespace.

## Reproduction

1. Create an agent.
2. Make namespaces.delete() fail, for example during a temporary backend outage.
3. Delete the agent with delete-backup-too=true.
4. Before this patch, the API returns 200 and the local agent is gone even though its memory namespace remains.

## Validation

- Added an API regression test covering the remote failure and retryable local state.
- uv run pytest -q — 299 passed, 24 skipped because live Moorcheh E2E requires a real key.
- uv run ruff check memanto/app/routes/sessions.py tests/test_api.py
- uv run ruff format --check memanto/app/routes/sessions.py tests/test_api.py

Bounty submission for #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Agent deletion with “delete backup too” now surfaces remote backup deletion failures instead of silently continuing.
  * Local agent state is preserved when the remote backup purge fails.
  * Unexpected deletion errors now return a consistent server error response, improving clarity for failed requests.

### DIFF
--- a/memanto/app/routes/sessions.py
+++ b/memanto/app/routes/sessions.py
@@ -176,8 +176,6 @@ async def delete_agent(
                 )
             )
         }
-    except AgentNotFoundError as e:
-        raise map_error_to_http_exception(e)
     except Exception as e:
         raise map_error_to_http_exception(e)
 
--- a/memanto/app/routes/sessions.py
+++ b/memanto/app/routes/sessions.py
@@ -158,14 +158,11 @@ async def delete_agent(
             )
 
         if delete_backup_too:
-            # Delete remote namespace only when explicitly requested.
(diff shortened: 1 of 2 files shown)


Neelpatel1604:
@dicnunz  Thank you for your contribution to Memanto and for taking the time to work on this bounty. We really appreciate the time and effort you put into your submission.

PR #1487, opened 2026-07-16
Fix as-of recall for memories that expired later

6pt6brty57-star:
## Problem

`search_as_of()` fetches memories through `_fetch_all_memories()`, which applies TTL against the current clock before the historical query can evaluate TTL at `as_of_date`. A memory that was valid at the requested point in time but expired later is therefore missing from historical recall.

Minimal timeline:

- created: `2020-01-01`
- requested as-of: `2020-06-01`
- expires: `2020-12-31`
- queried after expiration

Expected: the memory is returned because it was valid on June 1. Previous behavior: it was filtered out using today's clock.

## Fix

- Allow `_fetch_all_memories()` to skip current-time TTL enforcement when requested.
- Have `search_as_of()` fetch the unpruned records, then retain its existing expiration check against `as_of_date`.
- Keep current-time TTL enforcement unchanged for normal recall, recent recall, and changed-since queries.
- Add a regression test covering both a memory valid at the requested time and one already expired then.

## Validation

- `python -m pytest -q --disable-warnings --maxfail=1`
- `ruff check memanto/app/services/memory_read_service.py tests/test_memory_read_as_of_ttl.py`
- `ruff format --check memanto/app/services/memory_read_service.py tests/test_memory_read_as_of_ttl.py`
- `mypy memanto/app/services/memory_read_service.py`

Refs #770

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected historical memory searches to evaluate expiration times relative to the requested date.
  - Memories that were valid at the selected point in time are now returned accurately, even if their current expiration status differs.

- **Tests**
  - Added coverage verifying point-in-time expiration handling for historical searches.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -206,7 +206,12 @@ class MemoryReadService:
                     "temporal_mode": "as_of",
                 }
 
-            all_memories = self._fetch_all_memories(namespaces, type=type, tags=tags)
+            # Historical queries must evaluate TTL at the requested point in
+            # time. Applying the normal current-time TTL filter here would
+            # discard memories that were valid then but have expired since.
+            all_memories = self._fetch_all_memories(
+                namespaces, type=type, tags=tags, enforce_ttl=False
+            )
             all_memories = self._apply_temporal_filter(
                 all_memories, created_before=as_of_date
             )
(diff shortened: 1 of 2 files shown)


6pt6brty57-star:
Closing as a duplicate of #823. The earlier PR identified and fixed the same current-time TTL pruning in historical as-of recall. Apologies for missing it in the initial search.

PR #1495, opened 2026-07-16
fix(langgraph): upsert repeated store keys

Absamad-dew:
### Bounty submission

Addresses #770.

### Summary

Make `MemantoStore` preserve LangGraph's namespace/key upsert contract. A repeated `put()` for an existing key now updates the existing Memanto memory instead of creating another document with the same `lg:key:*` tag.

### Product bug and impact

LangGraph defines `put(namespace, key, value)` as “store or update an item”, with `key` serving as the unique identifier inside a namespace. `MemantoStore._do_put()` previously called `remember()` unconditionally, so this normal update sequence created two backend memories:

```python
store.put(("users",), "user-42", {"content": "Prefers email", "kind": "preference"})
store.put(("users",), "user-42", {"content": "Prefers Telegram", "kind": "preference"})
```

That can surface stale and current values together in search, grow storage on every profile/state update, and make Memanto behave differently from LangGraph stores that upsert by namespace/key.

### Fix

- Resolve the existing item through the integration's exact `lg:key:*` lookup.
- When a memory ID exists, call `SdkClient.update_memory()` with the new value, title, confidence, tags, source, and explicit type when supplied.
- Continue to call `remember()` for a genuinely new key.
- Preserve the existing search-cache invalidation behavior.

The patch intentionally does not delete historical duplicates already created by older versions; it avoids a destructive migration while preventing new duplicates in the normal sequential store flow.

### Regression coverage

`test_do_put_updates_existing_key_instead_of_creating_duplicate` provides an existing memory for the same namespace/key and verifies that:

- `update_memory()` is called with the existing memory ID and replacement fields;
- `remember()` is not called.

Before the patch the test fails because `update_memory()` is called zero times. After the patch it passes.

### Validation

- `.venv/Scripts/python.exe -m pytest integrations/langgraph/tests -q` — 38 passed.
- `.venv/Scripts/python.exe -m pytest -q` — 303 passed, 24 live E2E tests skipped because no real `MOORCHEH_API_KEY` is configured.
- `uv run ruff check .` — passed.
- `uv run ruff format --check .` — 247 files already formatted.
- `uv run mypy integrations/langgraph/langgraph_memanto/store.py` — passed.
- `git diff --check` — passed.

Refs #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updates for the same namespace/key now refresh the existing entry in place, preventing duplicate backend records.
  * Existing memory type is preserved when incoming updates omit type information.
  * Retrieval after an update reflects the latest written content.
  * Concurrent `put` calls for the same key within a store instance are serialized to avoid duplicates.
  * In strict mode, fallback recall failures now raise an error instead of silently returning no result.
  * Search cache entries for the affected namespace are invalidated after writes.
* **Tests**
  * Expanded coverage for update-vs-create behavior, omission of type fields, failure-path handling, and concurrency guarantees.


Absamad-dew:
Review update: I addressed both data-integrity findings in b361d97 and made the concurrency regression deterministic in b2b4155. CodeRabbit now reports no actionable comments, and both GitHub checks pass.

Local validation: 41 LangGraph integration tests; full suite 303 passed / 24 live E2E skipped; Ruff check/format, mypy, and git diff --check all pass.

The adapter now fails closed when the existence lookup fails and serializes same-key puts within one MemantoStore instance, including concurrent abatch dispatch. Backend-wide cross-process atomicity remains a server-side concern and is explicitly documented. Ready for maintainer review.

Xenogents:

PR #1498, opened 2026-07-17
Reject cross-site requests to local UI endpoints

silasbrookshaha:
## Summary
- Harden local-only UI management endpoints against cross-site browser requests.
- Allow same-origin localhost UI calls and non-browser local tooling, while rejecting remote Origin and cross-site Fetch Metadata headers.
- Add regression coverage for Origin, Sec-Fetch-Site, and loopback origin handling.

## Reproduction
1. Start the Memanto UI locally with the default wildcard CORS setting.
2. From another website, send a browser request to a local UI management endpoint such as /api/ui/shutdown.
3. Before this change, the endpoint only checked the TCP client address, so a browser-originated localhost request could pass the local guard.
4. After this change, cross-site browser requests are rejected with 403 while localhost same-origin requests still pass.

## Tests
- HOME=/private/tmp/memanto-test-home .venv/bin/python -m pytest tests/test_ui_auth.py tests/test_cors_fix.py -q
- HOME=/private/tmp/memanto-test-home .venv/bin/python -m pytest tests/test_remaining_ui_auth.py -q

Refs #770

## Summary by CodeRabbit

* **Security**
  * Strengthened protection for local UI management endpoints by rejecting cross-site browser requests and non-local `Origin` headers with HTTP 403.
  * Improved loopback detection by accepting `localhost` and common loopback/IP `Origin` formats, including IPv4-mapped IPv6.

* **Tests**
  * Added regression tests ensuring cross-site `Origin`/`Sec-Fetch-Site` scenarios are denied.
  * Added coverage verifying loopback `Origin` forms are accepted and remote origins are rejected.

### DIFF
--- a/memanto/app/ui/routes/ui_router.py
+++ b/memanto/app/ui/routes/ui_router.py
@@ -12,6 +12,7 @@ import signal
 import time
 from pathlib import Path
 from typing import Any
+from urllib.parse import urlparse
 
 from fastapi import (
     APIRouter,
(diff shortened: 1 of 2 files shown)


silasbrookshaha:
Thanks for the review. The loopback cross-site cases are already covered in this PR: both tests now route through an ASGI wrapper that presents a 127.0.0.1 client address, then assert the cross-site guard returns 403 with the expected detail.

I also re-ran the focused test locally:

```
pytest tests/test_ui_auth.py
16 passed, 1 warning
```

Xenogents:

PR #1502, opened 2026-07-17
fix: block cross-site management token theft

valencealignment:
## Summary

Prevent hostile browser pages and DNS-rebinding domains from inheriting Memanto's passwordless localhost management exemption. Explicitly credentialed management clients and legitimate localhost UI traffic, including cross-port frontends, continue to work.

Refs #770

## Flaw and impact

Memanto intentionally permits unauthenticated management requests from a loopback TCP client so the local CLI and browser UI remain convenient. The server also ships with `ALLOWED_ORIGINS=["*"]` and allows every CORS method and header.

A malicious website opened on the same machine could therefore call the primary `/api/v2` management API through `localhost`. Before this fix, the browser's connection appeared to originate from `127.0.0.1`, so the server trusted it without a management credential. The default CORS policy made the response readable to the attacking origin.

This allowed a site to:

1. create or enumerate agents;
2. activate an agent without a management key;
3. read the returned `session_token`;
4. use that token against the agent's memory read/write/delete endpoints.

The regression reproduced both critical steps before the fix:

- cross-site agent creation returned `201 Created`;
- cross-site activation returned `200 OK` with a live session token.

## Root cause

`require_management_access()` treated the socket peer's loopback address as sufficient authorization. That proves where the browser connection terminates, but not which web origin controls the browser. The exemption also did not validate the HTTP `Host`, so a DNS-rebinding domain could appear same-origin while resolving to the local service.

## Fix

Passwordless local management now requires all of the following:

- a loopback TCP peer;
- a loopback HTTP `Host` (`localhost`, IPv4/IPv6 loopback, with optional port);
- browser provenance that is not cross-site.

An explicit `Origin` is evaluated first. A loopback Origin is allowed even when a cross-port local frontend produces `Sec-Fetch-Site: same-site`; when Origin is absent, Fetch Metadata is used as the fallback. Remote origins, cross-site fetches, and non-loopback Host values are rejected.

Explicit API-key authorization remains available to remote clients and does not depend on the loopback exemption.

## Regression coverage

- rejects cross-site agent creation from a loopback socket;
- rejects cross-site activation and does not expose a session token;
- preserves same-origin localhost management access;
- preserves legitimate cross-port localhost UI access;
- rejects a DNS-rebinding Host header even when Fetch Metadata says `same-origin`.

## Validation

- focused security regressions: 5 passed
- full non-live test suite passed; 24 live Moorcheh E2E tests skipped without an API key
- Ruff check passed
- Ruff format check passed
- mypy passed
- `git diff --check` passed

/claim #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened management access controls for browser requests originating from outside the local environment.
  * Prevented cross-site requests from creating agents or obtaining session tokens through loopback access.
  * Preserved authorized management access for same-origin requests from localhost.

### DIFF
--- a/memanto/app/routes/auth_deps.py
+++ b/memanto/app/routes/auth_deps.py
@@ -118,14 +118,25 @@ def _is_loopback_origin(origin: str | None) -> bool:
     return parsed.hostname == "localhost" or _is_loopback_host(parsed.hostname)
 
 
+def _is_loopback_host_header(host: str | None) -> bool:
+    """Return True when an HTTP Host header names a loopback interface."""
+    if not host:
+        return False
+    try:
+        hostname = urlsplit(f"//{host}").hostname
+    except ValueError:
+        return False
+    return hostname == "localhost" or _is_loopback_host(hostname)
+
+
 def _is_cross_site_browser_request(request: Request) -> bool:
     """Detect browser requests that must not inherit loopback trust."""
-    fetch_site = request.headers.get("sec-fetch-site", "").strip().lower()
-    if fetch_site in {"cross-site", "same-site"}:
-        return True
-
     origin = request.headers.get("origin")
-    return origin is not None and not _is_loopback_origin(origin)
+    if origin is not None:
+        return not _is_loopback_origin(origin)
+
+    fetch_site = request.headers.get("sec-fetch-site", "").strip().lower()
+    return fetch_site in {"cross-site", "same-site"}
 
 
 def require_management_access(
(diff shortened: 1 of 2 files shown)


Xenogents:
Congratulations on winning the bounty @valencealignment! Of all the bounties we've been reviewing we've judged that your PR is one that points out the most severe security issue with Memanto. It is definitely not something we want for a website whether malicious or not, to that easily gain effectively full access to Memanto and its agentic capabilities.

We're still struggling at the moment on bountyhub side as their ui managed to break due to the sheer amount of claims being made but once that's complete we'll award the bounty to you and then eventually fully review and merge your PR. I believe your claim was made using a different PR you opened but I think it still works.

Thanks again for your contributions to Memanto and we welcome you to contribute to any other bounties!

PR #1503, opened 2026-07-17
fix(mcp): restrict automatic agent creation

valencealignment:
## Summary

- restrict MCP automatic agent creation to the configured default agent
- prevent ordinary memory tools from creating attacker-chosen agents when admin tools are hidden
- document the boundary and add regression coverage

## Root cause and impact

`MEMANTO_EXPOSE_ADMIN=false` removes the agent-management tools from the MCP surface, while `MEMANTO_AGENT_AUTO_CREATE` is documented as creating the configured default agent on first use. However, any memory tool call carrying a missing, non-default `agent_id` reached the same auto-create path. A prompt-injected or untrusted tool argument could therefore bypass the hidden administrative surface and create arbitrary agents/namespaces.

The lifecycle now auto-creates only `MEMANTO_DEFAULT_AGENT_ID`. Explicit non-default agents must already exist and otherwise fail with `AgentNotFoundError`.

## Validation

- reproduced the bypass with a regression test that failed before the fix
- MCP test suite: 27 passed
- Ruff check and format check passed
- mypy: success on 12 source files
- `git diff --check` passed

Refs #770

/claim #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented automatic creation of arbitrary agents when they are missing.
  * Missing non-default agents must now be created explicitly.
  * Preserved automatic creation and activation of the configured default agent.

* **Documentation**
  * Clarified the default-agent auto-creation setting and behavior for other agents.

### DIFF
--- a/integrations/mcp/README.md
+++ b/integrations/mcp/README.md
@@ -123,7 +123,7 @@ All config is via environment variables (load order: process env →
 | `MOORCHEH_API_KEY` | **yes** | — | Moorcheh API key. |
 | `MEMANTO_DEFAULT_AGENT_ID` | recommended | _none_ | Default agent. When set, tool calls may omit `agent_id`. |
 | `MEMANTO_AGENT_PATTERN` | no | `tool` | Pattern (`support`/`project`/`tool`) used when auto-creating the default agent. |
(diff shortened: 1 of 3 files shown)


Xenogents:

PR #1505, opened 2026-07-17
fix: temporal filter silently returns all memories on malformed timestamp

truongsontung:
## Summary of the flaw

`MemoryReadService._apply_temporal_filter` wrapped each filtering block in a broad `try/except (ValueError, AttributeError): pass`. Because the `pass` covered the *entire* list comprehension, **any** parse error disabled the filter and returned the full, unfiltered result set. This is a data-integrity bug in temporal queries (`created_after` / `created_before`):

1. **One bad record poisons the whole query** — if a single memory in the batch has an unparseable `created_at`, the exception aborts the comprehension and the filter returns *every* memory, including ones that should have been excluded.
2. **A malformed boundary is silently ignored** — passing an invalid `created_after`/`created_before` value returns all results instead of surfacing the error, silently over-returning data the caller asked to exclude.

## Reproduction

```python
svc = MemoryReadService(MagicMock())
results = [
    {'id': 'old', 'created_at': '2026-01-01T00:00:00Z'},
    {'id': 'bad', 'created_at': 'GARBAGE'},
    {'id': 'new', 'created_at': '2026-12-01T00:00:00Z'},
]
svc._apply_temporal_filter(results, created_after='2026-06-01T00:00:00Z')
# expected: [{'id': 'new'}]
# actual (before fix): all three records returned
```

## Fix

- Parse each record's `created_at` in isolation; an unparseable record is dropped (excluded) instead of aborting the whole filter.
- Treat a malformed caller-supplied boundary as a hard error (`ValueError`) rather than a silent no-op.
- Added `tests/test_memory_read_temporal_filter.py` covering correct filtering, the single-bad-record regression, and invalid-boundary handling.

All existing tests pass; 4 new regression tests added.

Relates to the Bug & Exploit Challenge (#770).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved temporal filtering so records with malformed `created_at` no longer cause the filter to fail for otherwise valid results.
  * Invalid `created_after`/`created_before` values now raise an error instead of being silently ignored.
* **Tests**
  * Added new test coverage for created-after/created-before filtering, including malformed record timestamps and invalid boundary inputs.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -583,23 +583,25 @@ class MemoryReadService:
             except (ValueError, AttributeError):
                 return None
 
-        filtered = results
+        if not created_after and not created_before:
+            return results
 
         # The boundary values are caller-supplied; a bad value is a hard error
         # rather than a silently-ignored filter (which would over-return data).
-        if created_after:
-            after_dt = parse_iso_timestamp(created_after)
-            filtered = [
-                r for r in filtered
-                if (c := _created(r)) is not None and c >= after_dt
-            ]
+        after_dt = parse_iso_timestamp(created_after) if created_after else None
+        before_dt = parse_iso_timestamp(created_before) if created_before else None
 
-        if created_before:
-            before_dt = parse_iso_timestamp(created_before)
-            filtered = [
-                r for r in filtered
-                if (c := _created(r)) is not None and c <= before_dt
-            ]
+        def _in_range(r: dict[str, Any]) -> bool:
+            c = _created(r)
+            if c is None:
+                return False
+            if after_dt is not None and c < after_dt:
+                return False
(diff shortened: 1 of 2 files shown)


truongsontung:
Thanks @coderabbitai — good call. Consolidated both bounds into a single pass so each record's `created_at` is parsed only once. Regression tests still pass.

truongsontung:
Hi — bumping this one too. This fixes a security-relevant fail-open in the temporal filter (`_apply_temporal_filter` bare `except` was passing through and returning ALL memories on a malformed timestamp). Added 4 regression tests, all green. Let me know if you'd prefer it split or adjusted. Thanks! 🙏

Xenogents:
Thanks for your contributions! This PR is being closed as the fixes have already been made by another PR.

PR #1489, opened 2026-07-16
Security/merge prs

Xenogents:
Merge remaining security-related PR's

- Remove url shell=true for to prevent potential command injections #774 
- Update remaining instances of datetime to use timezone #878 
- Update memory edit to overwrite memories instead of deletion and addition #878
- Support multi-tenant setups in langgraph #893 #1446
- Include # of rejected memories for batch counting response #1301
- Log config errors #1304
- Verify memory size for batch uploads #1304

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved per-agent session isolation with safe retries for recall/remember.
  * Updated remember retry behavior to trigger only on session-related errors.
  * Improved expiration handling for mixed naive/UTC datetimes.
  * Memory updates now overwrite existing records while preserving supported metadata.
  * Prevented unintended duplicate storage after write failures.
  * Configuration load failures now emit actionable warnings.

* **Improvements**
  * Standardized timezone-aware UTC timestamps across core and services.
  * Added batch submission content-length validation.
  * Improved batch write result reporting.
  * Made dashboard opening safer across supported platforms.

PR #1420, opened 2026-07-09
fix: rate limiting, timezone fixes, upsert update flow

ejspeed-cmd:
**Issue:** fixes #770  — Memanto Bug & Exploit Challenge  
## Summary

This PR resolves five bugs found during code review of the memory write and read pipeline. Changes span timezone correctness, rate-limit enforcement, a safer memory update flow, accurate batch result counting, and CRLF normalization for session summaries. A standalone probe script is included to validate all five issues against a live server.

---

### 1. Timezone-aware timestamps — `HIGH`

**Impact:** Runtime `TypeError` crash. Python raises on any comparison between a naive and an aware `datetime`, which happens during TTL expiry checks. Any agent using memory TTLs would hit an unhandled exception.

All `datetime.utcnow()` calls replaced with `datetime.now(timezone.utc)`.

Affected files: `core.py`, `memory_write_service.py`, `test_unit.py`

### 2. Rate limiting enforcement — `HIGH`

**Impact:** Security / availability. The rate-limiting module was fully implemented but never called, leaving all memory endpoints unprotected against abuse. A single agent or bad actor could flood write, read, delete and answer operations without any throttling.

Four guards wired in:

```
remember        → enforce_write_rate_limit
batch_remember  → enforce_write_rate_limit
edit_memory     → enforce_write_rate_limit
delete_memory   → enforce_delete_rate_limit
recall          → enforce_read_rate_limit
answer          → enforce_answer_rate_limit
```

### 3. Upsert-based memory update — `HIGH`

**Impact:** Permanent data loss. The old delete-then-upload flow would destroy a memory if the upload step failed after deletion — no recovery path. The new flow uploads first (Moorcheh treats same-ID uploads as upserts) and raises `MemoryError` on failure, leaving the original intact.

```mermaid
flowchart LR
    Old["Old flow\n(delete → upload)"] -->|"upload fails"| Loss["Memory lost permanently"]
    New["New flow\n(upsert in-place)"] -->|"upload fails"| Safe["Original preserved\nMemoryError raised"]
```

### 4. Batch result count fix — `MEDIUM`

**Impact:** Incorrect API response. `total_submitted` reported the input length rather than the number of records actually processed. When any item fails validation inside the service, the counts become inconsistent and callers cannot accurately determine how many memories were stored.

```python
# before
"total_submitted": len(memories)

# after
"total_submitted": len(results)
```

### 5. CRLF normalization in session summary — `MEDIUM`

[...]

PR #1507, opened 2026-07-17
fix(memory-write): preserve provenance and reject failed updates (#770)

evoesco:
## Summary

#1489 replaced the destructive delete-then-upload sequence with a same-ID overwrite. This focused follow-up closes two remaining integrity gaps in that path:

- content-only updates now preserve the stored `provenance`;
- failed, null, or missing backend statuses now raise `MemoryError` instead of returning `action: updated`;
- valid statuses remain case-insensitive through the existing `SUCCESSFUL_UPLOAD_STATUSES` contract.

It changes no public API and adds no delete, rollback, restore, or second-write path.

## Why it matters

On the original base (`5c3f06c`):

| Scenario | Current behavior | Expected behavior |
| --- | --- | --- |
| Edit a `validated` memory | Rewrites its provenance as `explicit_statement` | Preserve the stored attribution |
| Upload returns `failed`, `null`, or no status | Reports the memory as updated | Reject the unconfirmed update |

The first silently erases audit semantics. The second tells callers that an update persisted when the backend did not confirm it.

## Reproduction

```bash
python -m pytest -q tests/test_unit.py::TestMemoryWriteServiceUpdateIntegrity
```

With these regression tests applied to `5c3f06c`: **4 failed, 1 passed**.

On this PR: **5 passed**.

## Fix

- Carry the existing provenance through the same-ID replacement.
- Normalize upload statuses with `str(status).lower()` and the existing success-status set.
- Preserve specific `MemoryError` messages instead of wrapping them again.
- Cover the real `documents.get → MemoryReadService → update_memory` path plus `failed`, `null`, missing, and uppercase `OK` statuses.
- Keep the new test class distinct from the update tests added later on `main`, so both suites remain collectable after integration.

## Validation against latest main

Revalidated in a clean integration against `main` at `c4e3401`; the PR applies without conflicts, and GitHub reports it as mergeable.

- Both update test classes: **6 passed**
- Full suite: **488 collected; 464 passed, 24 skipped** (the skipped tests are live-key E2E)
- Ruff check: passed
- Ruff format check: passed (**259 files already formatted**)
- MyPy: passed (**78 source files**)
- `git diff --check`: passed

## Context and scope comparison

Earlier work in #1406 covered the broader update-safety path. After #1489 landed the same-ID overwrite, this PR was reduced to only the independently reproducible behaviors that remain missing.

Relative to #1385, this PR covers the same stored-provenance regression and also rejects unconfirmed backend outcomes (`failed`, null, and missing). #1385 additionally supports an explicit provenance override, so its provenance work is valid; #1507's advantage is combining provenance preservation with upload-outcome integrity in one focused patch.

Relative to #1420, this remains a narrow service-and-unit-test change that explicitly covers failed, null, missing, and case-insensitive successful statuses. #1420 addresses a broader set of issues and also rejects non-success outcomes, but its current update diff does not carry stored provenance; a null status reaches its outer error wrapper indirectly through `.lower()` rather than through the explicit normalization and regression coverage used here.

The practical advantage of #1507 is combined coverage in a small, directly reviewable patch—not a claim that the other PRs' separate fixes are invalid.

Related to #770.

## Social / technical discussion

A public technical write-up of the residual update-integrity failure modes, reproduction, and validation is available in the current r/MachineLearning self-promotion thread (which explicitly permits project posts):

- https://www.reddit.com/r/MachineLearning/comments/1ul5bgf/comment/oyh2vpm/

### DIFF
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -327,6 +327,7 @@ class MemoryWriteService:
                 confidence=updates.get("confidence", metadata.get("confidence", 0.8)),
                 status=updates.get("status", metadata.get("status", "active")),
                 tags=updates.get("tags", metadata.get("tags", [])),
+                provenance=metadata.get("provenance") or "explicit_statement",
             )
 
             # Update timestamps (preserve created_at, set updated_at to now)
@@ -382,16 +383,24 @@ class MemoryWriteService:
             except Exception as e:
                 raise MemoryError(f"Upload failed. Error: {e}")
 
+            status = upload_result.get("status", "unknown")
+            if str(status).lower() not in SUCCESSFUL_UPLOAD_STATUSES:
+                raise MemoryError(
+                    f"Failed to upload updated memory {memory_id}: {status}"
+                )
+
             return {
                 "id": memory_id,
                 "namespace": namespace,
-                "status": upload_result.get("status", "unknown"),
+                "status": status,
                 "action": "updated",
                 "reason": "Memory updated successfully via overwrite",
                 "validation": validation_result.get("action", "validated"),
                 "updated_fields": list(updates.keys()),
             }
 
+        except MemoryError:
+            raise
         except Exception as e:
             raise MemoryError(f"Failed to update memory: {e}")
 
--- a/tests/test_unit.py
+++ b/tests/test_unit.py
@@ -468,7 +468,7 @@ class TestMemoryWriteServiceDelete:
         assert "validation_count" not in uploaded
 
 
-class TestMemoryWriteServiceUpdate:
+class TestMemoryWriteServiceUpdateIntegrity:
     def test_update_memory_preserves_read_service_metadata(self):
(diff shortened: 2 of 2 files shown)


evoesco:
Hi @het0814 — I'm updating this existing review request rather than adding a second ping before the Aug 1 bounty cutoff.

#1507 remains a focused two-file follow-up after #1489. It preserves stored provenance on content-only same-ID updates and rejects failed, null, or missing backend statuses instead of reporting an unconfirmed update as successful.

Fresh validation against `main` at `c4e3401`: clean integration; 488 tests collected, 464 passed, and 24 live-key E2E tests skipped; 6 focused tests passed; Ruff check and format check passed; MyPy passed across 78 source files.

GitHub reports the PR as mergeable, CodeRabbit is green, and the TypeScript SDK workflow is still waiting for maintainer approval. I also updated the PR body with a neutral scope comparison against #1385 and #1420.

Could you please approve the waiting workflow and review when convenient? I'm ready to address any requested changes.

PR #1511, opened 2026-07-17
fix: fix: temporal filter silently returns all memories on malfor

key1989han:
Closes #1505

fix: temporal filter silently returns all memories on malformed timestamp

/claim #1505

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the temporal filter could return all memories when provided with a malformed timestamp.
  * Added documentation for the fix related to issue 1505.


Xenogents:
This is a duplicate.

PR #1513, opened 2026-07-17
fix: fix: block cross-site management token theft

key1989han:
Closes #1502

fix: block cross-site management token theft

/claim #1502


Xenogents:
This is a duplicate.

PR #1517, opened 2026-07-18
fix(security): rate limiter fail-closed for unknown operations (issue #1438)

### LINKED ISSUE #1438: [BOUNTY] 🐜 Rate limiter fail-open + 2 validation inconsistencies

truongsontung:
## Summary
- Fixes #1438: `RateLimiter.check_rate_limit()` was **fail-open** — any operation not present in `self.limits` (typos, unknown ops, newly added routes) was silently allowed with no rate limiting at all.
- Now unknown operations fall back to a conservative default limit (`30 req / 60s`) instead of being allowed unlimited.
- `enforce_rate_limit()` detail message now references the resolved `limit` to avoid a `KeyError` for unknown operations.

## Test plan
```bash
python -m pytest tests/test_rate_limiting.py -q
```
Added regression tests:
- unknown op is capped by the default limit (not unlimited)
- unknown op returns a positive `retry_after` once the window is exhausted
- known ops (`health`, `memory_answer`) still behave correctly

## Impact
Closes a rate-limit bypass that let unrecognized operations evade throttling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rate-limit enforcement for previously unconfigured operations.
  * Requests now receive a conservative default limit instead of bypassing rate limiting.
  * Rate-limit responses provide accurate retry timing and limit details.
  * Existing configured operations continue to enforce their specified limits.

### DIFF
--- a/memanto/app/utils/rate_limiting.py
+++ b/memanto/app/utils/rate_limiting.py
@@ -38,6 +38,11 @@ class RateLimiter:
             "health": RateLimit(300, 60),  # 300/min
         }
 
+        # Fail-closed default: any operation not explicitly configured is still
+        # rate-limited instead of silently allowed (prevents bypass via typos /
+        # unknown operation names). Chosen conservatively strict.
+        self.default_limit = RateLimit(30, 60)
+
     def _get_key(self, operation: str, agent_id: str) -> str:
         """Generate rate limit key"""
         return f"{operation}:{agent_id}"
@@ -49,10 +54,9 @@ class RateLimiter:
         Check if request is within rate limit
         Returns: (allowed, retry_after_seconds)
         """
-        if operation not in self.limits:
-            return True, None
-
-        limit = self.limits[operation]
+        # Fail-closed: unknown operations fall back to a safe default limit
+        # rather than being silently allowed (was a fail-open bypass).
+        limit = self.limits.get(operation, self.default_limit)
         key = self._get_key(operation, agent_id)
         now = time.time()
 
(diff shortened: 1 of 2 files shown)


Xenogents:
Thanks for your contributions! We're closing this PR as these fixes have already been covered by another PR.

PR #1534, opened 2026-07-20
fix: serialize session renewal and logout

nexicturbo:
## Summary

Serializes each agent's session lifecycle so parallel auto-renewals cannot mint competing replacement tokens and a renewal already in flight cannot publish a fresh bearer token after logout, while unrelated agents remain concurrent.

Addresses #770.

## Flaw and impact

Session validation, renewal, creation, termination, and publication of the global active-session marker were separate unsynchronized operations.

### Competing renewals

When two requests for the same agent arrived inside the renewal window, both could observe the same near-expiry session and both call `renew_session()`. Each renewal creates a different JWT and overwrites the same session JSON and global active-session marker. Only the last token written remains valid. This can:

- return an already-invalid replacement token or cookie to one caller;
- intermittently lock a browser or parallel SDK client out on its next request; and
- raise `FileNotFoundError` when renewals race while unlinking the shared `active` marker.

### Logout bypass race

`end_session()` did not share the renewal lock. If logout overlapped a request already inside `check_and_auto_renew()`, this ordering was possible:

1. Renewal reads the old active session and passes the near-expiry check.
2. Logout marks that old session terminated and returns successfully.
3. The in-flight renewal creates a new session, persists a new JWT, and marks it active.

The result is a fresh, valid bearer token created after logout completed. An already authenticated/stolen client racing the victim's logout can therefore retain access. The opposite write order can also return a token that is invalid immediately.

This is a concurrent execution / time-of-check-to-time-of-use failure (CWE-362/CWE-367) at the session-revocation boundary.

## Reproduction

Four deterministic regression tests cover the lifecycle guarantees:

- `test_auto_renew_is_single_flight_per_agent` forces two requests into the near-expiry path together and verifies exactly one replacement token is issued.
- `test_end_session_revokes_concurrent_auto_renewal` pauses a renewal after eligibility is established, starts logout, then releases renewal. Logout must wait, terminate the newly issued session, and make that fresh JWT fail validation.
- `test_lifecycle_operations_for_different_agents_can_overlap` pauses one agent's session write and verifies another agent can complete session creation independently.
- `test_active_session_read_waits_for_marker_update` verifies readers cannot inspect the process-wide active marker while it is being replaced.

Before the lifecycle fix, logout completes against the old session while renewal is paused; release then publishes a new active token and the revocation assertion fails.

## Fix

- Use one reentrant lifecycle lock per agent for session creation, termination, and the read/check/renew sequence.
- Use a separate narrow reentrant lock for every read, replacement, and deletion of the process-wide active-session marker.
- Keep token rotation reentrant so `check_and_auto_renew()` can call the normal `renew_session()` / `create_session()` path without deadlock, without blocking unrelated agents during session file I/O.
- Ensure a concurrent logout is authoritative: if renewal started first, logout terminates the replacement; if logout started first, renewal observes the terminated session and does nothing.

## Validation

- Full local suite: **366 passed, 24 skipped** (**390 collected**)
  - All skips are live Moorcheh E2E tests gated on a real `MOORCHEH_API_KEY`
- All four focused concurrency regressions: passed
- Ruff lint: passed
- Ruff format check: passed
- Mypy on the changed service: passed
- `git diff --check`: passed

## Checklist

- [x] Deterministic regression tests added
- [x] Existing unit and API tests pass
- [x] Ruff lint and formatting checks pass
- [x] Mypy passes for the changed service
- [x] No credentials or secrets added

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved session lifecycle reliability during simultaneous creation, renewal, termination, and deletion.
  - Prevented conflicting operations from producing invalid or inconsistent active-session and token states.
  - Ensured automatic renewal is single-flight per agent under concurrent checks.
  - Improved consistency of the active-session marker, avoiding partial/mid-update reads.

- **Tests**
  - Expanded concurrency test coverage across same and different agents.
  - Added assertions verifying concurrent lifecycle actions correctly invalidate and revoke tokens.

### DIFF
--- a/memanto/app/services/session_service.py
+++ b/memanto/app/services/session_service.py
@@ -9,6 +9,7 @@ import json
 import logging
 import os
 import secrets
+import threading
 from datetime import timedelta
 from pathlib import Path
 from typing import Any
@@ -70,6 +71,13 @@ class SessionService:
             or self._generate_secure_secret_key()
         )
         self.secret_key: str = resolved_secret_key
+        self._auto_renew_locks: dict[str, threading.RLock] = {}
+        self._auto_renew_locks_guard = threading.Lock()
+
+    def _get_auto_renew_lock(self, agent_id: str) -> threading.RLock:
+        """Return the per-agent lock that serializes token rotation."""
+        with self._auto_renew_locks_guard:
+            return self._auto_renew_locks.setdefault(agent_id, threading.RLock())
 
     def _generate_secure_secret_key(self) -> str:
         """Generate (or reuse) a persisted fallback secret for JWT signing.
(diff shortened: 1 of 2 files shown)


nexicturbo:
@coderabbitai review

nexicturbo:
Social amplification for the #770 scoring matrix: https://x.com/TurboNexic/status/2079102310848897380

nexicturbo:

PR #1539, opened 2026-07-20
fix(app): decouple conflict detection prompt from embedded query to prevent context overflow

### LINKED ISSUE #1329: `detect-conflicts` (on-prem) fails on any active day because the conflict query exceeds the embedding model's context window
## Environment

- `memanto 0.2.4`, on-prem backend brought up via the Docker quick start (`moorcheh up`), all defaults, zero config.
- Embedding model: `nomic-embed-text` (context length 2048); LLM: `qwen2.5` — the quick-start defaults.
- Two Compose containers on a shared network: `moorcheh-onprem-server` (`:8080`) and `moorcheh-ollama` (`:11434`).

This is the documented out-of-the-box configuration, so the failure reproduces with the quick start and no custom settings.

## Current behavior

`memanto detect-conflicts` fails for any day with a non-trivial amount of session activity:

```
Failed to detect conflicts: Conflict detection failed: ollama embeddings request returned error status
```

The mechanism: `DailyAnalysisService.generate_conflict_report` (`memanto/app/services/daily_analysis_service.py`) concatenates that day's session summaries into `full_text`, folds them into a single `conflict_prompt`, and passes the whole thing as the retrieval `query` to `client.answer.generate(..., top_k=50)`. The on-prem server embeds that query through Ollama. With the default embedding model `nomic-embed-text` (context length 2048 tokens) from the quick start, and the embedding request issued without truncation, any query longer than the context window is rejected.

Reproduced on `memanto 0.2.4`, on-prem backend, `nomic-embed-text` embedding. The failure is purely a function of query length. Hitting the on-prem `/search` endpoint (the same server-side embedding path) with increasing slices of a real session summary:

```
chars=20     HTTP 200
chars=2000   HTTP 200
chars=4000   HTTP 200
chars=6000   HTTP 400  {"message":"ollama embeddings request returned error status","status":"error"}
chars=11190  HTTP 400  {"message":"ollama embeddings request returned error status","status":"error"}
```

Confirming the cause at the Ollama layer, with real (dense) text over 2048 tokens:

```
POST /api/embed      {"input": <6k chars>}                    -> HTTP 200 (prompt_eval_count=2048, silently truncated)
POST /api/embed      {"input": <6k chars>, "truncate": false} -> HTTP 400 "the input length exceeds the context length"
POST /api/embeddings {"prompt": <6k chars>}                   -> HTTP 500 "the input length exceeds the context length"
```

`remember`, `recall`, and small-query `answer` are unaffected, since their queries stay well under 2048 tokens. The consequence is that conflict detection breaks precisely on the days with enough activity to be worth checking; a single day's ~11 KB summary is already well past the limit.

## Expected behavior

`detect-conflicts` should complete regardless of how much session content a day accumulates, and produce a conflict report (empty or not). A day's volume of memories should never surface as an embedding error.

## Suggested approaches

Roughly from narrowest to most structural:

1. Bound the retrieval query. The query used to pull historical memories does not need to be the entire day's text; a context-sized window or a compact digest is enough to retrieve the relevant memories, and avoids the limit. Note that `/answer` uses `query` for both the retrieval embedding and the LLM prompt, so this should decouple the two, moving the instruction block into `header_prompt`/`footer_prompt` while a shorter digest serves as the embedded `query`.
2. Chunk the day and detect per chunk. Split the session content into context-sized chunks, retrieve per chunk, and union the candidate historical memories before the LLM pass. Higher recall than (1), at the cost of more embedding calls.
3. Truncate at the embedding call. Have the on-prem embedding path request Ollama with `truncate: true` (the `/api/embed` default) rather than letting the request fail. This prevents the hard error, but silent truncation to the first 2048 tokens is lossy and would mask the design issue that (1) and (2) address.

Approaches (1) and (2) belong in `memanto`; (3) is a complementary server-side guard in the Moorcheh on-prem embedding path, not a substitute. Separately, the error surfaced to the user could name the cause (query length versus context window) rather than the opaque upstream status. Let's discuss the preferred shape.

sirEven:
### Root Cause

`generate_conflict_report()` in `daily_analysis_service.py` stuffed the entire instruction prompt plus the full day's session content into the `query` parameter of `client.answer.generate()`. The `query` is embedded for similarity retrieval, so any input exceeding the embedding model's context window (2048 tokens for the default `nomic-embed-text`) caused a hard error:

```
Failed to detect conflicts: Conflict detection failed: ollama embeddings request returned error status
```

This breaks precisely on days with enough session activity to be worth checking — a single day's ~11 KB summary is already well past the limit.

### Fix (Approach #1 from the issue)

Decouple the LLM instructions from the embedded query:

1. **`header_prompt`**: Contains the conflict analysis instructions + full session content. Passed to the LLM but NOT embedded for retrieval.
2. **`footer_prompt`**: Contains the JSON response format instructions. Also NOT embedded.
3. **`query`**: Contains a truncated digest of the session content (max 6000 chars ~ 1500 tokens), well within the 2048-token embedding context window. Used for similarity retrieval only.

The `answer.generate()` API already supports `header_prompt` and `footer_prompt` (used by the answer endpoint and conversation extraction service), so this is a natural use of existing infrastructure rather than new chunking logic.

### Testing

6 new regression tests in `tests/test_conflict_query_length.py`:

1. **test_conflict_report_query_stays_within_context_window** — verifies query stays under 6000 chars even with 11KB+ session content
2. **test_conflict_report_uses_header_and_footer_prompts** — verifies instructions are in header/footer, not query
3. **test_conflict_report_query_does_not_contain_instruction_text** — verifies CRITICAL INSTRUCTIONS etc. are NOT in the query
4. **test_conflict_report_short_content_still_works** — verifies the fix works correctly with normal-sized content
5. **test_conflict_report_preserves_ai_model_kwarg** — verifies ai_model is still passed when set
6. **test_conflict_report_no_ai_model_when_unset** — verifies ai_model is omitted when not set (backward compat)

Full test suite: 368 passed, 24 skipped, 0 failures. Ruff, mypy, and format checks all pass.

### Observed but out of scope

`generate_summary()` has the same pattern (full_text in query, no header/footer prompts) and would fail the same way for days with large session content. This could be addressed in a follow-up issue.

Part of the Bug and Exploit Challenge (#770) bounty on BountyHub.

Closes #1329

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved conflict report generation for sessions with extensive content.
  - Prevented oversized retrieval requests that could exceed context limits.
  - Separated session content from report instructions for more reliable processing.
  - Preserved AI model selection behavior when configured or unavailable.
  - Maintained successful report generation for shorter session content.

- **Tests**
  - Added coverage for query length limits, prompt handling, instruction isolation, and model configuration scenarios.


Xenogents:

sirEven:
> Thank you for your contribution to Memanto and for taking the time to work on this bounty. We really appreciate the time and effort you put into your submission.
> 
> The bounty winner has already been determined but we will still continue to be merging useful bug fixes into the repository.
> 
> Thanks again for contributing, and we hope to see more of your contributions to Memanto!

Where do you publish those winners?

Xenogents:
We post about it on our youtube channel, video for this bounty is here https://www.youtube.com/watch?v=0VZPl3xeD_I. Not a very standardized format but we let people know a little bit about how we've decided on winners.

sirEven:
> We post about it on our youtube channel, video for this bounty is here https://www.youtube.com/watch?v=0VZPl3xeD_I. Not a very standardized format but we let people know a little bit about how we've decided on winners.

Honest critique: 
I suggest you guys introduce verifiable reports on your bounties, that give contributors the ability of understanding your decisions. Even more so when you keep merging solutions that are not being taken into account when granting those bounties.

Xenogents:
Sorry if my wording seemed like we determined a winner first and then started merging all these solutions, we have looked through everyone's changes before deciding on a winner. But yes I agree there isn't a whole lot of transparency into our decision-making partly because of an influx of bots creating hundreds of PR's causing us to forego personalized feedback to try to run through the backlog as fast as possible.

We've now implemented a system which puts users through a captcha to restrict bot activity so hopefully we won't have to deal with that again. I think your suggestion is a good idea though so I'll run it by my team and we can hopefully improve transparency for future bounties. Thanks for letting us know and open to any other critiques you have.

PR #1540, opened 2026-07-20
fix(okf): preserve literal entry delimiters during round trips

nexicturbo:
## Summary

- prevent literal `<!-- okf-entry -->` text in memory bodies or YAML metadata from becoming stacked-file framing
- reversibly HTML-escape every string in only affected documents and record the encoding in `x_memanto`
- decode affected bodies and frontmatter after the loader has safely split real records
- add a deterministic two-memory regression covering body, title, tags, resource, and pre-existing HTML entities

Refs #770

## Root cause and impact

OKF uses the literal HTML comment `<!-- okf-entry -->` to separate records in stacked files. The exporter writes memory content and metadata into each document, and the loader calls `text.split(ENTRY_DELIMITER)` over the complete Markdown text before parsing either frontmatter or body.

A user memory containing that ordinary text is therefore treated as multiple records during the supported Memanto → OKF → Memanto flow. In the regression, two exported memories become three imported memories: the affected memory is truncated at the marker, its remaining content becomes a phantom untyped record, and its metadata no longer describes the complete memory. The same collision affects both stacked and per-memory files because the loader splits every file.

This is realistic for memories that preserve documentation, templates, source snippets, or discussions of the OKF format itself. More generally, user-controlled memory text must not be able to alter the export framing.

## Reproduction

The added test exports two facts with `split="type"`. One body contains:

```text
Document the literal internal marker & keep &amp; and <tags> unchanged:
<!-- okf-entry -->
It is part of the memory, not a record boundary.
```

Before the fix:

```text
FAILED tests/test_okf.py::test_round_trip_preserves_literal_entry_delimiter_in_memory_content
AssertionError: assert 3 == 2
```

The first memory is truncated and a third phantom memory is created.

## Fix

When a document body or any nested frontmatter string contains the framing sentinel, the exporter applies reversible `html.escape(..., quote=False)` encoding to every string in that document and records `x_memanto.content_encoding = "html-escaped-v1"`. Encoding the complete affected document also prevents collisions in title, description, tags, resource, and Memanto metadata while preserving pre-existing entities such as `&amp;` exactly. The loader reverses body and frontmatter encoding only after real entries have been separated. Unaffected exports remain byte-for-byte unchanged and human-readable.

## Validation

- `python -m pytest tests/test_okf.py -q` — 6 passed
- `python -m pytest -q` — 363 passed, 24 skipped (live API key tests)
- `python -m ruff check memanto/app/services/okf_export_service.py memanto/cli/migrate/okf_loader.py tests/test_okf.py`
- `python -m ruff format --check memanto/app/services/okf_export_service.py memanto/cli/migrate/okf_loader.py tests/test_okf.py`
- `python -m mypy memanto/app/services/okf_export_service.py memanto/cli/migrate/okf_loader.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved literal stacked-entry delimiter text in exported and re-imported content.
  * Prevented delimiter-like text in document content and metadata from corrupting OKF imports.
  * Preserved nested metadata, tags, resources, and HTML entities during round trips.
* **Tests**
  * Added coverage confirming neighboring entries remain correctly separated while content is preserved.

### DIFF
--- a/memanto/app/services/okf_export_service.py
+++ b/memanto/app/services/okf_export_service.py
@@ -21,6 +21,7 @@ Memanto -> OKF -> Memanto round-trips keep them. OKF consumers ignore unknown
 frontmatter keys.
 """
 
+import html
 import re
 import shutil
 from datetime import datetime
@@ -36,6 +37,7 @@ from memanto.app.utils.validation import validate_output_path, validate_safe_id
 # the loader can split them back apart without colliding with ``---`` that may
 # appear inside a document body (e.g. the migrate ``[Supporting data]`` footer).
 ENTRY_DELIMITER = "<!-- okf-entry -->"
+CONTENT_ENCODING_HTML_ESCAPED = "html-escaped-v1"
 
 # Default: collapse a type into a single stacked file once it exceeds this many
 # memories (see the ``auto`` split mode).
@@ -275,6 +277,7 @@ class OkfExportService:
     def _render_okf_doc(self, mem: dict[str, Any], mem_type: str) -> str:
         """Render a single memory dict as one OKF markdown document."""
         content = (mem.get("content") or "").strip()
+        rendered_content = content
         title = mem.get("title") or "Untitled"
 
         frontmatter: dict[str, Any] = {"type": mem_type, "title": title}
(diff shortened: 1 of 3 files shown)


nexicturbo:
Social amplification for the #770 scoring matrix: https://x.com/TurboNexic/status/2079323856179769381

nexicturbo:

PR #1473, opened 2026-07-15
[Bounty] Bug report: Memory write service - data integrity audit

zhangjianming7051:
## Summary

This PR adds a detailed bug bounty report to \docs/bounty_reports/\ identifying **4 findings** (1 critical, 2 medium, 1 low) in the memory write service.

### Critical: Non-atomic update causes silent data loss
\update_memory()\ deletes the old version before uploading the new one. If the upload fails (network error, server timeout, quota), the memory is permanently destroyed with no rollback path.

### Medium: Timezone-aware/naive datetime inconsistency  
\MemoryRecord.created_at\ defaults to naive \datetime.utcnow()\ but \update_memory()\ parses stored ISO strings as aware datetimes. Comparison between the two raises \TypeError\.

### Medium: Missing import timestamp validation
Imported memories bypass all timestamp sanity checks — far future dates, \created_at > updated_at\, and improper timezone conversions pass through silently.

### Low: Batch operation partial failure ambiguity
\atch_store_memories()\ can silently succeed for only a subset of items without clear error signalling to the caller.

## Bounty Claim
This submission is under the \ Memanto Bug & Exploit Challenge (Issue #770).

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive markdown bounty report detailing seven audit findings related to memory write reliability, deletion-success handling, pagination correctness, and inconsistent timestamp handling (naive vs timezone-aware).
  * The report includes severity classifications and an updated summary table covering each finding.

PR #1300, opened 2026-06-30
Prevent failed batch writes from creating phantom summaries

2y999:
## Summary

Fixes a data consistency bug in batch memory writes. When Moorcheh batch upload returned a non-success status such as `error` or `unknown`, pending items were not counted as failed. The API/CLI batch paths then logged every submitted `MemoryRecord` into the local session Markdown summary even though those memories were not actually accepted by the backend.

This creates "phantom memories": daily summaries, conflict scans, and migration reporting can later treat failed batch writes as real stored memories because they appear in the local session summary.

This PR:
- treats non-success batch upload statuses as per-item failures
- only writes successful batch items (`queued`, `success`, `ok`) to session summaries
- applies the same summary guard to API batch writes, conversation-extraction writes, and DirectClient/CLI batch writes
- adds regression coverage for service counts, API summary logging, and DirectClient summary logging

## Validation

- `python -m pytest tests/test_unit.py::TestMemoryWriteServiceBatch tests/test_unit.py::TestForgetEndToEnd::test_failed_batch_does_not_write_session_summary tests/test_api.py::TestMEMANTOAPI::test_failed_batch_remember_does_not_log_phantom_summary -q`
- `python -m pytest tests/test_unit.py -q`
- `python -m pytest tests/test_api.py -q`
- `python -m ruff check memanto/app/services/memory_write_service.py memanto/app/routes/memory.py memanto/cli/client/direct_client.py tests/test_api.py tests/test_unit.py`
- `python -m ruff format --check memanto/app/services/memory_write_service.py memanto/app/routes/memory.py memanto/cli/client/direct_client.py tests/test_api.py tests/test_unit.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory write success is now determined by normalized write/upload statuses (`queued`, `success`, `ok`); unexpected statuses mark affected items as failed with clearer per-item error details.
  * Single and batch memory writes no longer create “phantom” session Markdown notes; session Markdown logging runs only for successful items, aligned per-item in batch flows.
  * API responses more accurately reflect write outcomes by gating success/failure counts and returning the appropriate status when writes fail.

PR #1542, opened 2026-07-20
Memory integrity/merge prs

Xenogents:
- Prevented failed batch memory writes from improperly generating "phantom" session summaries. #1300 
- Prevent tag metadata leaking into memory content, string/datetime conversion errors #1328
- Fix text parsing when content contains "Tags:", fixed a temporal filter bug where a single bad timestamp would disable the filter entirely #1440
- Replace UTC with timezone aware datetimes, ensures created_at <= updated_at, throws a ValueError when limit+offset > 100 #1473 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session expiration and duration calculations for consistent UTC handling.
  * Memory write results now accurately report successful and failed items.
  * Prevented failed memory writes from appearing in session summaries.
  * Improved parsing of memory content and trailing tag metadata.
  * Preserved imported timestamps and supported string expiration values.
  * Prevented malformed timestamps from disrupting temporal filtering.

* **Tests**
  * Added regression coverage for failed batch writes, timestamp handling, tag parsing, and expiration metadata.


het0814 (approved):
LGTM!

PR #1483, opened 2026-07-16
fix: replace stale OKF export snapshots

valencealignment:
## Summary

Make repeated OKF exports exact snapshots instead of overlays. The exporter now renders a complete bundle in a sibling staging directory and swaps it into place only after every section and index has been written successfully.

Refs #770

## Root cause and impact

`OkfExportService.write_okf_bundle()` previously wrote directly into an existing `{agent_id}_okf` directory without clearing it. If a memory was deleted, renamed, moved to another type, or changed between split layouts, its old Markdown file remained in the bundle.

A later `load_okf_bundle()` import scanned those stale files and treated them as current memories. That could resurrect deleted knowledge or duplicate renamed memories, breaking the bundle's memory integrity. A write failure could also leave a partially refreshed cache.

## Reproduction

1. Export an agent with `Old fact` and `Old event` using `split="file"`.
2. Export the same agent to the same default path with only `Current fact`.
3. Before this fix, `memories/fact/old-fact.md` and the entire `memories/event/` directory remain.
4. Importing the refreshed bundle returns all three entries even though the second snapshot contains only one.

The new regression test demonstrates this exact sequence.

## Fix

- Render each export into a uniquely named sibling staging directory.
- Move the previous bundle aside only after the new snapshot is complete.
- Swap the completed snapshot into the requested path.
- Restore the previous bundle if the final rename fails.
- Remove temporary and backup directories after the swap.
- Add regression coverage for stale-entry removal and failed-export preservation.

## Validation

- `uv run pytest -q` — 335 passed, 24 skipped (live Moorcheh E2E tests require an API key)
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv run mypy .`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * OKF bundle exports now publish as a complete, refreshed snapshot, removing stale files and leftover per-type content.
  * Failed exports no longer corrupt the previously working bundle; it stays readable.
  * Bundle reads (including single-file reads) no longer catch intermediate replacement states; loads wait safely.
  * Temporary staging/backup artifacts are cleaned up automatically, even after errors.
  * Missing-bundle errors now point to the user-facing path for clearer troubleshooting.

PR #1552, opened 2026-07-20
fix(okf): replace stale bundle content on re-export

nexicturbo:
## Summary

- make each OKF re-export a snapshot of current agent state
- remove only Memanto-managed bundle sections before rebuilding them
- prevent deleted memories, old context documents, and obsolete split layouts from surviving on disk
- add a regression covering memory deletion, file-to-type layout changes, removed context, and an empty final export

Refs #770

## Root cause and impact

`write_okf_bundle` creates the output directory with `exist_ok=True`, then overwrites only paths needed by the current export. It never removes paths produced by an earlier export.

That means deleting a memory and exporting the same agent again leaves its old Markdown file in `memories/`. `load_okf_bundle` recursively imports every Markdown file in that section, so the supposedly deleted memory is resurrected on the next import. Switching from per-file to stacked layout is worse: both layouts remain and the same live memories can be imported twice. Daily summaries, session logs, and metrics can also remain after they no longer exist in current state.

This is a deterministic memory-integrity failure in the supported export/import workflow, not a cosmetic artifact.

## Reproduction

1. Export facts `Keep` and `Deleted` to the default `agent1_okf` path with `split=file`.
2. Re-export the same agent and path with only `Keep`.
3. Load the resulting bundle with `load_okf_bundle`.

Before the fix, the loader returns both records:

~~~text
[(Deleted, memories/fact/deleted.md),
 (Keep, memories/fact/keep.md)]
~~~

The regression also switches the second export to `split=type`, proving obsolete per-memory files do not coexist with the new stacked file.

## Fix

Before writing a bundle, clear only the paths owned by the exporter: the root index plus `memories`, `daily-summaries`, `sessions`, and `metrics`. Symlink and file cases are unlinked without traversal; managed directories are removed and rebuilt. Unknown files at the bundle root are preserved.

The new test verifies that deleted memories and removed context disappear, layout changes do not duplicate data, and an empty re-export leaves no importable memories or stale metrics.

## Validation

- `uv run pytest tests/test_okf.py` — 6 passed
- `uv run pytest` — 363 passed, 24 skipped (live API-key E2E tests)
- `uv run ruff check memanto/app/services/okf_export_service.py tests/test_okf.py`
- `uv run ruff format --check memanto/app/services/okf_export_service.py tests/test_okf.py`
- `uv run mypy memanto/app/services/okf_export_service.py`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Re-exporting an OKF bundle now removes outdated memories, summaries, sessions, metrics, and index content.
  - Reusing an existing export location now accurately reflects the latest available content.
  - Empty exports no longer retain stale folders or files from previous exports.

- **Tests**
  - Added coverage to verify that re-exported bundles replace removed content correctly.

### DIFF
--- a/memanto/app/services/okf_export_service.py
+++ b/memanto/app/services/okf_export_service.py
@@ -107,6 +107,7 @@ class OkfExportService:
             assert validated is not None
             base = validated
         base.mkdir(parents=True, exist_ok=True)
+        self._clear_managed_output(base)
 
         # Section order mirrors how an agent would read the bundle.
         sections: dict[str, str] = {}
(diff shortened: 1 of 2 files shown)


nexicturbo:
Closing as a duplicate of #1483, which already covers stale OKF re-export snapshots with a stronger transactional replacement. Thanks.

PR #1553, opened 2026-07-20
fix: serialize concurrent session summary writes

nexicturbo:
## Summary
- serialize session-summary header creation and appends through the shared SessionService
- build each memory/deletion entry before entering the critical section and write it as one record
- add a deterministic eight-worker regression for the first-write race

## Bug
Memory API routes dispatch summary logging through asyncio.to_thread. Previously, every worker checked whether the summary existed and then issued many independent append writes. Concurrent first writes could all observe a missing file, duplicate its header, and interleave record fields. These summaries feed daily analysis and conflict detection, so the race can corrupt the local history even when the remote memory writes succeeded.

The regression synchronizes eight concurrent first-existence checks against unpatched main. It fails there with five duplicated headers in the reproduced run; the same test passes with this patch.

## Tests
- uv run ruff check memanto/app/services/session_service.py tests/test_session_summary_concurrency.py
- uv run mypy memanto/app/services/session_service.py
- uv run pytest -q (all tests pass; 24 live E2E tests skipped because no real Moorcheh API key is configured)

## Social field report
- https://x.com/TurboNexic/status/2079352402595995801

### DIFF
--- a/memanto/app/services/session_service.py
+++ b/memanto/app/services/session_service.py
@@ -11,6 +11,7 @@ import os
 import secrets
 from datetime import timedelta
 from pathlib import Path
+from threading import Lock
 from typing import Any
 
 import jwt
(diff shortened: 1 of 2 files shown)


nexicturbo:

PR #1556, opened 2026-07-21
fix: return union for multi-type recall

nexicturbo:
## Summary by CodeRabbit
* **Bug Fixes**
  * Improved memory recall when multiple memory types are requested by running type-specific searches and merging results.
  * Preserved relevance ranking by re-ranking combined results and removing duplicates.
  * Prevented duplicate backend searches when the same memory type is requested multiple times.
  * Added validation so invalid memory types are rejected before any search is executed.
* **Tests**
  * Added regression coverage for multi-type recall, deduplication, validation, and concurrent execution behavior.
 

## Social field report

https://x.com/TurboNexic/status/2079420719004291469

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -113,17 +113,31 @@ class MemoryReadService:
             if not namespaces:
                 return {"results": [], "total_found": 0, "execution_time": 0}
 
-            # Build enhanced query with Moorcheh metadata filters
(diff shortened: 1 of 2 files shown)


nexicturbo:
Withdrawing this separate bounty PR to comply with #770's one-PR rule. Its implementation and regression coverage are preserved in the active consolidated submission #1571.

PR #1562, opened 2026-07-21
fix: enforce Python client query guards

nexicturbo:
## Summary

Enforce Memanto's existing query-size and result-count guards in both Python client implementations.

Addresses #770.

## Flaw and impact

The REST recall and answer endpoints reject questions over 1,000 characters and bound retrieval counts to 1–100. `DirectClient` and `SdkClient` did not enforce the same contract:

- recall accepted and forwarded oversized queries;
- answer accepted and forwarded oversized questions;
- answer accepted malformed or unbounded `limit` values such as `0`, `101`, `1.5`, `"10"`, and `True`.

That bypassed Memanto's local cost guard and allowed invalid `top_k` values or embedding/RAG inputs to cross the backend boundary, producing avoidable backend failures and potentially expensive requests.

## Root cause

Both clients already shared `_validate_query()` for recall, but that helper checked only blank input and the recall limit. The answer paths performed only their own blank-question check and never called the shared limit validator.

## Fix

- Extend `_validate_query()` in both clients to enforce `InputLimits.MAX_QUERY_LENGTH`.
- Route both answer implementations through that same validator.
- Add deterministic coverage for DirectClient and SdkClient across recall and answer.
- Prove that all rejected inputs fail before either storage search or Moorcheh answer generation is called.

## Validation

- Complete non-live test suite passed; 24 live Moorcheh E2E tests skipped because no real API key is configured.
- 14 focused regression cases passed.
- Ruff check passed.
- Ruff format check passed.
- MyPy passed for both client modules.
- `git diff --check` passed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added validation to reject queries exceeding the maximum allowed length.
  * Standardized validation for questions and result limits across client implementations.
  * Invalid inputs are now rejected before requests are sent to underlying services.

### DIFF
--- a/memanto/cli/client/direct_client.py
+++ b/memanto/cli/client/direct_client.py
@@ -1191,8 +1191,7 @@ class DirectClient:
         # Ensure there is a valid, non-expired session for this agent
         session = self._get_validated_session_for_agent(agent_id)
 
-        if not question or not question.strip():
-            raise ValueError("Question must be a non-empty string")
+        self._validate_query(question, limit)
 
         # get namespace from session
         namespace = session.namespace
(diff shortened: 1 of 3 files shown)


nexicturbo:
Social field report: https://x.com/TurboNexic/status/2079441152999084270

nexicturbo:

PR #1569, opened 2026-07-21
fix(claudecode): distill only the current Stop turn

nexicturbo:
## Summary

Claude Code invokes the configured `Stop` hook whenever Claude finishes a response. The integration previously sent the cumulative transcript tail to Memanto on every turn, so durable decisions from earlier turns were repeatedly extracted and stored until they aged out of the tail.

This change distills only the user turn associated with the current `Stop` event. It also anchors asynchronous hook reads with `last_assistant_message`, preventing a delayed hook from ingesting a newer turn appended after the hook started.

Bounty submission for #770.

## Root cause and impact

The hook treated `Stop` as a once-per-session event, but Claude Code documents it as an event that runs when the agent finishes responding. The existing `stop_hook_active` guard only identifies continuation caused by a Stop hook; it does not suppress normal Stop events on later user turns.

As a session grew, each turn re-submitted earlier user and assistant messages to `distill_and_store()`. That could:

- create duplicate or near-duplicate durable memories;
- make retrieval noisier by overweighting old decisions;
- repeatedly spend LLM tokens and backend writes on unchanged history; and
- let an async hook accidentally process a later turn instead of the turn that spawned it.

Official hook semantics:

- https://code.claude.com/docs/en/hooks
- https://code.claude.com/docs/en/hooks-guide

## Reproduction

1. Create a Claude Code transcript with two completed user/assistant turns.
2. Run the existing Stop-hook transcript reader after the second turn.
3. Observe that the returned text contains both the first turn's durable decision and the second turn, so the first decision is submitted again.
4. For the async race, append a third turn before the second turn's hook reads the file. The old reader consumes the third turn even though it belongs to a later Stop event.

The added regression tests model both cases. Before this fix, the first case includes `Old decision: use unittest`, and there is no way to anchor the second Stop event before the appended third turn.

## Fix

- Parse transcript messages once into role/text pairs.
- Anchor the readable snapshot to the Stop event's `last_assistant_message`.
- Slice from the latest user message before that anchor through the anchored assistant response.
- Retain the most recent user-invoked `/skill` up to the anchor for memory tagging.
- Preserve `read_transcript_text()` behavior for callers that intentionally need a bounded cumulative tail.

## Validation

- `python -m pytest integrations/claudecode/tests -q --disable-warnings` — 70 passed
- `python -m pytest -q --disable-warnings` — full non-live suite passed; 24 live-key E2E tests skipped as expected
- `python -m ruff check integrations/claudecode/hooks/_common.py integrations/claudecode/hooks/on_stop.py integrations/claudecode/tests/test_hooks.py`
- `python -m ruff format --check integrations/claudecode/hooks/_common.py integrations/claudecode/hooks/on_stop.py integrations/claudecode/tests/test_hooks.py`
- `python -m mypy integrations/claudecode/hooks/_common.py integrations/claudecode/hooks/on_stop.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transcript handling for varied message formats and role labels.
  * Distillation now focuses on the specific conversation turn that triggered the stop hook.
  * Skill detection is limited to the relevant turn, preventing older or unrelated commands from being selected.
  * Added safeguards to exclude content from earlier and later turns.

* **Tests**
  * Added regression coverage for turn scoping and assistant-message anchoring.

### DIFF
--- a/integrations/claudecode/hooks/_common.py
+++ b/integrations/claudecode/hooks/_common.py
@@ -130,52 +130,71 @@ def read_transcript_text(
     find (string content, or content blocks carrying a ``text`` field),
     labelling it by role when available. Returns the trailing ``max_chars``.
     """
-    _, rendered = _read_transcript_full(
-        transcript_path, max_messages=max_messages, max_chars=max_chars
-    )
-    return rendered
+    messages = _read_transcript_messages(transcript_path)
+    return _render_messages(messages, max_messages=max_messages, max_chars=max_chars)
 
 
 def read_transcript_for_distillation(
     transcript_path: str | None,
+    last_assistant_message: str | None = None,
     max_messages: int = 40,
     max_chars: int = 8000,
 ) -> tuple[str | None, str]:
-    """Return ``(skill, tail_text)`` from a single pass over the transcript.
+    """Return ``(skill, current_turn_text)`` for one Stop-hook invocation.
 
-    Long sessions truncate to the tail for the LLM (the latest discussion is
-    where decisions usually crystallise), but the user's original ``/tdd`` or
-    ``/grill-with-docs`` invocation typically sits at the very start of the
-    conversation and would fall outside that tail. We scan the entire
-    transcript for the first skill token, then return it alongside the
-    truncated text so ``distill_and_store`` can tag memories correctly even
-    on long sessions.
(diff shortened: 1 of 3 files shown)


nexicturbo:

PR #1570, opened 2026-07-21
fix(mcp): isolate concurrent agent sessions

nexicturbo:
## Summary

- keep one independently activated `SdkClient` per agent in the MCP lifecycle
- route all seven memory and RAG tools through their agent-scoped client while retaining the shared client for agent administration
- add deterministic concurrency coverage plus a same-agent client reuse test

## Problem

`SdkClient` stores its active agent and session as mutable instance state. The MCP lifecycle previously shared one client across every agent and held its lock only during `ensure_ready`. This allowed the following interleaving:

1. request A activates the shared client for `agent-a` and releases the lifecycle lock
2. request B activates that same client for `agent-b`
3. request A calls an SDK operation for `agent-a`, but the client now carries `agent-b`'s session

The SDK then rejects request A with a `SessionError`. Every MCP memory tool was exposed to this race in multi-agent servers.

## Fix

The lifecycle now lazily creates and activates one session client per agent. A new `client_for(agent_id)` method returns the stable scoped client, and each memory/RAG tool keeps that returned client through its SDK operation. The existing `client` property remains the administrative client used by agent CRUD tools.

## Validation

- `uv run --group dev --with-editable integrations/mcp pytest integrations/mcp/tests -q` — 27 passed
- `uv run --group dev pytest -q` — 374 passed, 24 live E2E tests skipped because external credentials were not configured
- `uv run --group dev ruff check .`
- `uv run --group dev ruff format --check .`
- `uv run --group dev mypy .`
- `git diff --check`

Contribution for #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved isolation for simultaneous operations across different agents by maintaining separate active sessions per agent.
  * Prevented one agent’s session activation from affecting another agent’s session.
  * Ensured repeated operations for the same agent reuse the same active session reliably.
  * Updated agent-scoped tools to consistently acquire the correct per-agent session before making requests.
* **Tests**
  * Added coverage for independent per-agent sessions, session reuse, and concurrent initialization behavior.

### DIFF
--- a/integrations/mcp/memanto_mcp/lifecycle.py
+++ b/integrations/mcp/memanto_mcp/lifecycle.py
@@ -10,8 +10,9 @@ to call ``recall`` and expects it to "just work". This module owns:
   underlying ``SdkClient._get_validated_session_for_agent`` already auto-renews
   tokens nearing expiry, so each client only has to be activated once.
 
-All operations are guarded by a lock so concurrent tool invocations don't race
-to create the same agent twice.
+The client registry is guarded by a short-lived global lock, while per-agent
+locks serialize setup for the same agent without blocking unrelated agents on
+network calls.
 """
 
 from __future__ import annotations
@@ -44,6 +45,7 @@ class MemantoLifecycle:
         self._settings = settings
         self._client = SdkClient(api_key=settings.api_key_value())
         self._session_clients: dict[str, SdkClient] = {}
+        self._agent_locks: dict[str, threading.Lock] = {}
         self._ensured_agents: set[str] = set()
         self._lock = threading.Lock()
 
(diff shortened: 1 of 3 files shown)


nexicturbo:
Review-ready update: commit `c090a99` addresses the initialization-serialization feedback. Client/lock registry updates remain under the short-lived global lock, while `get_agent` / auto-create / activation network work runs under an agent-specific lock, so unrelated agents initialize concurrently and same-agent callers remain serialized.

Fresh validation on the current PR head:
- `uv run --group dev --with-editable integrations/mcp pytest integrations/mcp/tests/test_lifecycle.py -q` — 3 passed
- `uv run --group dev ruff check integrations/mcp/memanto_mcp/lifecycle.py integrations/mcp/tests/test_lifecycle.py` — passed
- `git diff --check` — passed

The barrier-based regression `test_different_agents_initialize_in_parallel` verifies the cross-agent concurrency behavior. The PR is mergeable and ready for maintainer review.

nexicturbo:
Withdrawing this separate bounty PR to comply with #770's one-PR rule. Its implementation and regression coverage are preserved in the active consolidated submission #1571.

PR #1582, opened 2026-07-21
fix(ui): reject cross-site localhost requests

nexicturbo:
## Summary

The Web UI management guard trusted only the connection's client IP. That does not establish that a browser request originated from the Memanto UI: when a hostile website submits a form to `127.0.0.1`, the victim browser connects to Memanto from a loopback socket and passes the existing guard.

This change requires the complete local-browser boundary: a loopback peer, a localhost Host, non-cross-site Fetch Metadata, and a matching Origin when present. Headerless local CLI requests remain supported.

Bounty submission for #770.

## Root cause and impact

Every `/api/ui/*` management route depends on `_require_local`, including shutdown, on-prem restart, configuration/API-key changes, filesystem browsing, integration installation, migrations, summaries, and conflict resolution. `_require_local` previously checked only `request.client.host`.

A hostile page could therefore use a simple, no-preflight HTML form to send:

```http
POST /api/ui/shutdown HTTP/1.1
Host: 127.0.0.1:8000
Origin: https://attacker.example
Sec-Fetch-Site: cross-site
Content-Type: application/x-www-form-urlencoded
```

The request was accepted with HTTP 200. In UI mode, that route schedules `SIGINT` against the Memanto process. The bodyless on-prem restart route was exposed through the same cross-site form primitive. A DNS-rebinding origin also passed because the guard did not validate Host.

## Reproduction

The new regression recreates the browser boundary with an ASGI client whose network peer is `127.0.0.1`, then sends the hostile Origin/Fetch Metadata headers above. Before this fix:

```text
200 {"status":"ignored","reason":"Not in UI mode"}
```

The `ignored` response is expected only because the deterministic test leaves UI mode disabled; it proves the request reached the privileged handler. With UI mode enabled, the handler queues process termination.

## Fix

- Keep the existing loopback peer check.
- Require the HTTP Host to be an IP loopback address, `localhost`, or a reserved `.localhost` name.
- Reject `Sec-Fetch-Site` values other than `same-origin` and `none`.
- Canonicalize and compare HTTP(S) Origin scheme, hostname, and effective port.
- Continue allowing local CLI clients that omit browser-only Origin and Fetch Metadata headers.
- Add regressions for a cross-site form POST, DNS rebinding, valid same-origin UI traffic, and headerless local CLI traffic.

## Validation

- `python -m pytest tests/test_ui_auth.py tests/test_remaining_ui_auth.py tests/test_api.py -q --disable-warnings` — passed
- `python -m pytest -q --disable-warnings` — full non-live suite passed; 24 credential-dependent live tests skipped
- `python -m ruff check .`
- `python -m ruff format --check .`
- `python -m mypy memanto/app/ui/routes/ui_router.py`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened protection for local-only UI management endpoints by validating the request host, origin, and browser fetch metadata.
  * Blocked cross-site and DNS-rebinding-style requests, including attempts using loopback-resolving hostnames.
  * Preserved access for same-origin loopback browser requests and local command-line usage.
* **Tests**
  * Added coverage for loopback, cross-origin, IPv4-mapped IPv6, and missing-header scenarios.

### DIFF
--- a/memanto/app/ui/routes/ui_router.py
+++ b/memanto/app/ui/routes/ui_router.py
@@ -12,6 +12,7 @@ import signal
 import time
 from pathlib import Path
 from typing import Any
+from urllib.parse import urlsplit
 
 from fastapi import (
     APIRouter,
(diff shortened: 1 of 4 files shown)


nexicturbo:
Closing this public PR to comply with issue #770's instruction that security vulnerabilities be reported privately first. The validated report and patch are being submitted through the repository's private vulnerability-reporting channel for coordinated review.

PR #1602, opened 2026-07-22
fix: 5 bugs from Memanto Bug & Exploit Challenge (#770)

ochsec:
fix: 5 bugs from Memanto Bug & Exploit Challenge (#770)

Bugs 1-3: Timeline amnesia in temporal query methods
- search_as_of, search_changed_since, search_recent all called
  _fetch_all_memories() which unconditionally filtered out currently-expired
  memories using datetime.now(). This broke point-in-time recall for any
  memory that had since expired.
- Fix: Added skip_ttl_filter parameter to _fetch_all_memories(). Temporal
  methods now pass skip_ttl_filter=True so they apply their own time-relative
  expiry logic without pre-filtering.

Bug 4: Query string filter injection
- _build_filtered_query() concatenated user query directly into the Moorcheh
  search string alongside #key:value metadata filters with no escaping.
  A user could inject #memory_type:fact into the query to bypass
  caller-specified type=["preference"] filters.
- Fix: Strip # from user query text before concatenation.

Bug 5: update_memory cannot modify expired memories
- update_memory() called get_memory() which filtered out expired memories,
  raising "Memory not found" - making it impossible to extend TTL, edit
  content, or revive expired memories.
- Fix: Added include_expired parameter to get_memory() (default False).
  update_memory() now calls get_memory() with include_expired=True.

All 380 existing unit tests pass. Reproduction scripts confirm all 5 bugs
are fixed.

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving expired memories when explicitly requested.
  * Temporal searches now correctly evaluate memories based on their historical or relative time conditions.

* **Bug Fixes**
  * Expired memories can now be updated instead of being treated as missing.
  * Improved query handling to prevent filter injection through search text.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -55,8 +55,13 @@ class MemoryReadService:
             self._namespace_service = NamespaceService(self.client)
         return self._namespace_service
 
-    def get_memory(self, memory_id: str, namespace: str) -> dict[str, Any] | None:
-        """Retrieve specific memory by ID with TTL enforcement"""
+    def get_memory(self, memory_id: str, namespace: str, include_expired: bool = False) -> dict[str, Any] | None:
+        """Retrieve specific memory by ID with TTL enforcement.
+
+        When include_expired is True, expired memories are returned without
+        TTL filtering. This is needed by update_memory to allow reviving or
+        modifying expired memories (e.g. extending TTL, editing content).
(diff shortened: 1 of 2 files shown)


Xenogents:
Thanks for your contributions! The query string filter injection is an intentional feature and memory expiration is more of a stale feature for Memanto.

PR #1615, opened 2026-07-23
fix: return renewed header session token

spoconymacius3879254-ctrl:
## Summary

Header-authenticated API clients can lose authentication during automatic session renewal. The server revokes the old session and creates a replacement, but previously returned that replacement only through a cookie. A client using `X-Session-Token` therefore receives a successful renewal-triggering response and then gets a 401 on its next request, with no way to learn the new credential.

This change:

- returns the renewed credential in `X-Session-Token` when the request used header authentication;
- exposes that response header through CORS so browser API clients can read it; and
- adds regression coverage for the full old-token/replacement-token lifecycle and CORS exposure.

The raw token is not logged or placed in a response body. Cookie-authenticated behavior is unchanged.

## Reproduction

1. Activate an agent and authenticate only with `X-Session-Token`.
2. Trigger renewal by placing the session inside `SESSION_EXTEND_THRESHOLD_MINUTES`.
3. Make one authenticated request. It succeeds and revokes the original session.
4. Before this fix, the response provides no replacement header, so the next header-authenticated request fails with 401.

The added test forces the renewal threshold, asserts the replacement header is present, confirms the old token is rejected, and confirms the returned token authenticates successfully.

## Verification

- `pytest`: 390 passed, 24 live-backend E2E tests skipped because they require external credentials
- `ruff check memanto/app/main.py memanto/app/routes/auth_deps.py tests/test_api.py tests/test_cors_fix.py`: passed

Submission for #770.

## Summary by CodeRabbit

* **New Features**
  * Automatically renews near-expiry authenticated sessions using a session token provided via request header.
  * Returns replacement session tokens to clients in the `X-Session-Token` response header.
* **Bug Fixes**
  * Ensures renewed sessions use the replacement token rather than a revoked one.
  * Keeps cookie- and header-authenticated sessions working after renewal.
* **Testing**
  * Added coverage to confirm the custom `X-Session-Token` header is exposed to browser clients via CORS.

### DIFF
--- a/memanto/app/main.py
+++ b/memanto/app/main.py
@@ -97,6 +97,9 @@ app.add_middleware(
     allow_credentials=settings.CORS_ALLOW_CREDENTIALS,
     allow_methods=["*"],
     allow_headers=["*"],
+    # Header-authenticated API clients must be able to read an auto-renewed
+    # token from the response. Custom response headers are not CORS-safelisted.
+    expose_headers=["X-Session-Token"],
 )
 
 # Include routers
--- a/memanto/app/routes/auth_deps.py
+++ b/memanto/app/routes/auth_deps.py
@@ -226,6 +226,12 @@ def get_current_session(
             # fails signature/session_id validation.
             if session_cookie:
                 set_session_cookie(response, renewed.session_token, request)
+            # API clients commonly authenticate with X-Session-Token instead
+            # of the browser cookie.  Hand them the replacement credential as
+            # well; otherwise auto-renewal revokes their only token and their
+            # next request is necessarily rejected.
+            if x_session_token:
+                response.headers["X-Session-Token"] = renewed.session_token
 
         return session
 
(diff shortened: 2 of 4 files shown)


spoconymacius3879254-ctrl:
Addressed the review nit in da930ee: the regression now clears the activation cookie before constructing header auth, so the old-token rejection and replacement-token success paths are genuinely header-only. Reverified: targeted test passes, full suite 390 passed / 24 credentialed live E2E skipped, and Ruff passes.

freeEasyLoope:
**Suggested fix** — 2026-07-23 22:05

```
```json
{
  "type": "code",
  "language": "JavaScript (Node.js/Express example)",
  "code": "// Middleware or route handler for session renewal\napp.post('/api/session/renew', async (req, res) => {\n  const oldSessionToken = req.headers['x-session-token'];\n  \n  if (!oldSessionToken) {\n    return res.status(401).json({ error: 'No session token provided' });\n  }\n\n  try {\n    // Validate and r
```

Xenogents:

PR #1616, opened 2026-07-23
fix(langgraph): preserve namespace tuple boundaries

evoesco:
## Summary

- prevent structurally distinct LangGraph namespace tuples from collapsing onto the same Memanto agent
- preserve unambiguous legacy agent IDs while using a versioned, reversible representation when the legacy mapping is ambiguous
- decode versioned IDs when listing namespaces without misinterpreting legacy IDs that merely resemble the new marker

## Impact

The integration relies on a namespace tuple to select the backend Memanto agent. The current mapping is not one-to-one:

```python
("team_alpha", "project")
("team", "alpha_project")
```

Both become:

```text
langgraph_team_alpha_project
```

Because `_ensure_client` pools and activates clients by that derived `agent_id`, reads, writes, and searches from both structural namespaces are routed to the same backend agent.

In applications that use namespaces to separate workflows—or encode tenant or customer boundaries in namespace segments—this can cause cross-workflow or cross-tenant memory mixing, unintended disclosure during recall/search, and wrong-context reads or writes that logically corrupt the application's memory state. This is a deterministic collision, not a probabilistic hash collision.

## Root cause

The integration derived `agent_id` with `"_".join(namespace)` and reconstructed namespaces with `split("_")`. Delimiters inside a namespace segment were therefore indistinguishable from tuple boundaries.

## Fix

- Use the legacy representation only when every namespace segment is unambiguous.
- Encode ambiguous namespaces as versioned URL-safe Base64 over canonical JSON UTF-8.
- Validate decoded values by re-encoding them before accepting the versioned interpretation.
- Keep the empty namespace, ordinary legacy namespaces, and marker-like legacy IDs backward compatible.
- Route `_ensure_client` and namespace listing through the same centralized mapping helpers.

## Compatibility

The empty namespace and simple legacy namespaces retain their current IDs. Namespaces containing underscores, Unicode, empty segments, the singleton `("default",)`, or the version marker use the reversible representation.

Existing memories already stored under a previously ambiguous legacy ID are not migrated automatically. Correct isolation requires distinct IDs for those namespaces going forward.

## Current branch status and #1317

The same root collision was identified independently in #1317. GitHub records that PR as merged into `integrations/merge-prs`, but at the current branch tips its exact head (`b185984`) is not an ancestor of either `main=c4e3401` or `integrations/merge-prs=ccb192e`. Both live branches still contain the original `"_".join(namespace)` mapping, so the collision remains reachable in the code maintainers would ship today.

This PR targets `main` directly. It also preserves ordinary unambiguous legacy agent IDs and switches only ambiguous namespaces to the reversible form. This is a current-branch reachability and compatibility distinction, not a claim that #1317's diagnosis was invalid.

## Validation

- Focused store suite: **25 passed**
- Full LangGraph integration suite: **48 passed**
- Ruff check: passed on both changed files
- Ruff format check: passed on both changed files
- MyPy: passed on `integrations/langgraph/langgraph_memanto/store.py`
- `git diff --check`: passed
- CodeRabbit: successful, with no actionable review comments

Related to #770.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved namespace-to-agent mapping to prevent distinct namespaces from being conflated.
  * Preserved support for legacy namespace mappings while enabling safe handling of empty, Unicode, and complex namespace segments.
  * Improved namespace listing so encoded namespaces are accurately restored, including default and empty namespaces.
  * Ensured memory operations consistently target the correct namespace-specific agent.

### DIFF
--- a/integrations/langgraph/langgraph_memanto/store.py
+++ b/integrations/langgraph/langgraph_memanto/store.py
@@ -8,7 +8,7 @@ Mapping between abstractions
 ----------------------------
 
     BaseStore                         ->  Memanto
-    namespace (tuple[str, ...])       ->  agent_id       (``langgraph_<p0>_<p1>...``)
+    namespace (tuple[str, ...])       ->  agent_id       (legacy or reversible ID)
     key (str)                         ->  reserved tag   ``lg:key:<key>``
     value["kind"] / value["type"]     ->  memory_type    (auto-parsed if absent)
     value["title"]                    ->  title          (auto-derived if absent)
@@ -34,7 +34,11 @@ Documented limitations
 from __future__ import annotations
 
 import asyncio
+import base64
+import binascii
+import json
 import logging
+import re
 import threading
 import time
 from collections.abc import Iterable
@@ -59,6 +63,8 @@ logger = logging.getLogger(__name__)
 _KEY_TAG_PREFIX = "lg:key:"
 _ENCODED_KEY_TAG_PREFIX = "lg:key:v1:"
 _RESERVED_PREFIX = "lg:"
+_NAMESPACE_ENCODING_MARKER = "v1_"
+_LEGACY_NAMESPACE_PART = re.compile(r"[A-Za-z0-9-]+")
 
 _VALID_MEMORY_TYPES = {
     "fact",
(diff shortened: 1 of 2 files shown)


evoesco:
Hi @het0814 and @Xenogents — could one of you review #1616 when convenient?

One approving review from a maintainer with write access is the remaining merge-policy gate. The code head is unchanged at `99ecfae`.

The current mapping is non-injective: `("team_alpha", "project")` and `("team", "alpha_project")` both address `langgraph_team_alpha_project`, so structurally distinct namespaces share one Memanto agent. Where namespaces represent workflow or tenant boundaries, that can mix memories, expose one workflow's recall results to another, and produce wrong-context reads or writes.

GitHub records #1317 as merged to `integrations/merge-prs`, but its exact head is not an ancestor of the current `main` or current integration branch; both live branches still retain the vulnerable join mapping. #1616 targets `main` directly and preserves ordinary legacy IDs while encoding only ambiguous namespaces.

Fresh verification on `99ecfae`: 25 focused tests and all 48 LangGraph integration tests passed; Ruff check, Ruff format check, MyPy, and `git diff --check` passed. I'm ready to address any feedback.

Xenogents:

We've decided that rather than preventing namespace collision, it's a better user experience to not use base64 hashes in namespaces as to not confuse them.

PR #1617, opened 2026-07-23
fix(read): search_as_of recalls since-expired memories (bounty #770)

StreetTriple765999:
# [Bounty #770] Fix search_as_of timeline amnesia: since-expired memories silently dropped

## Summary of the flaw

`search_as_of` answers a **point-in-time** question: "What was true at this date?" It should return every memory that was **created on/before `as_of_date` and not yet expired at `as_of_date`**.

However, it delegates fetching to `MemoryReadService._fetch_all_memories`, which **always runs `_filter_expired_memories` against the current wall-clock time** before returning. As a result, any memory whose `expires_at` falls **between `as_of_date` and now** (valid at the target date, but already expired today) is removed **before** `search_as_of` ever gets to evaluate it.

The net effect is **timeline amnesia / recall accuracy loss**: a TTL-bearing memory can never be recalled for the historical window in which it was actually live. This matches the High-severity criteria in the bounty scope ("loses track of when an event occurred (timeline amnesia)", "fails to recall highly relevant facts").

The code itself acknowledges the risk of "timeline amnesia / poor recall" in a comment near `search_memories`, yet `search_as_of` still inherits the current-time expiry filter through `_fetch_all_memories`.

## Reproduction steps

1. Create a memory with:
   - `created_at = 2026-01-10` (before the target date)
   - `expires_at = 2026-06-01` (after the target date, but before "now")
2. Query `search_as_of(as_of_date="2026-01-15", ...)`.
3. **Before fix:** the memory is **absent** (dropped by `_filter_expired_memories` because it is expired *now*), even though it was live on 2026-01-15.
4. **After fix:** the memory is correctly returned.

A self-contained regression test is in `tests/test_as_of_expired_recall.py` (fake client, no backend/API key). On unpatched code it fails with `timeline amnesia: since-expired memory lost`; on patched code it passes.

Minimal PoC (see tests/test_as_of_expired_recall.py for the full case):

    service.search_as_of(as_of_date="2026-01-15", agent_id="agent-1", limit=None)
    # expected: "temporal-fact" (created 01-10, expires 06-01) IS recalled
    # actual (bug): it is missing

## Fix / architectural proposal

1. Add a `filter_expired: bool = True` parameter to `_fetch_all_memories`.
2. When `filter_expired` is `False`, skip the current-time `_filter_expired_memories` pass.
3. `search_as_of` now calls `_fetch_all_memories(..., filter_expired=False)` and relies on its **existing** `expires_at <= as_of_dt` check, which already correctly excludes only memories that had expired *by the target date*.

Other callers (`search_changed_since`, `search_recent`) keep the default `filter_expired=True` (unchanged) - for "recent"/"changed since" queries, dropping currently-expired memories is desired; for a **past point-in-time** query it is not.

## Impact

- **Severity: High** - silently corrupts point-in-time recall for any TTL-based memory; the entire purpose of `search_as_of` ("what was true then?") is defeated for expiring memories.
- **Reproducibility: deterministic** - pure logic flaw, reproduced with a fake client (no network/backend), consistent across runs.
- Backward compatible: no API/schema changes; only `search_as_of` behaviour changes (only toward *more correct* recall).

## Verification

- New regression test `tests/test_as_of_expired_recall.py` passes after the fix and fails before it.
- Existing `test_memory_read_as_of.py`, `test_memory_read_temporal_recall.py`, `test_memory_read_confidence.py`, `test_memory_read_filter_sanitization.py` still pass (no regression).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved point-in-time (“as of”) memory searches to correctly include items that were valid at the requested date, even if they expired afterward.
  * Fixed delete-and-recreate cases to consistently return the historical version that existed at the selected time.
  * Enhanced handling of memory expiration when `expires_at` is stored as a datetime, preventing incorrect exclusions/keeps.

* **Tests**
  * Added regression coverage for datetime-typed expiration values in “as of” queries, alongside existing expiration and delete-and-recreate scenarios.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -237,7 +237,9 @@ class MemoryReadService:
                     "temporal_mode": "as_of",
                 }
 
-            all_memories = self._fetch_all_memories(namespaces, type=type, tags=tags)
+            all_memories = self._fetch_all_memories(
+                namespaces, type=type, tags=tags, filter_expired=False
+            )
             all_memories = self._apply_temporal_filter(
                 all_memories, created_before=as_of_dt.isoformat()
             )
@@ -426,6 +428,7 @@ class MemoryReadService:
         namespaces: list[str],
         type: list[str] | None = None,
         tags: list[str] | None = None,
+        filter_expired: bool = True,
     ) -> list[dict[str, Any]]:
         """
         List all stored memories across the given namespaces via Moorcheh's
@@ -434,6 +437,13 @@ class MemoryReadService:
 
         Iterates through all pages using cursor-based pagination (next_token)
         so results are not truncated at the 100-item per-page cap.
+
+        ``filter_expired`` controls whether memories expired at the current
+        wall-clock time are dropped. Point-in-time callers such as
+        ``search_as_of`` must pass ``filter_expired=False`` and apply their
+        own expiry check against the target date, otherwise memories that
+        were valid at that past date but have since expired are silently
+        dropped (timeline amnesia).
         """
         items: list[Any] = []
         for ns in namespaces:
@@ -479,7 +489,9 @@ class MemoryReadService:
 
             memories.append(formatted)
 
-        return self._filter_expired_memories(memories)
+        if filter_expired:
+            return self._filter_expired_memories(memories)
+        return memories
 
     def _memory_version_key(
         self, memory: dict[str, Any], fetch_index: int
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -238,7 +238,11 @@ class MemoryReadService:
                 }
 
             all_memories = self._fetch_all_memories(
-                namespaces, type=type, tags=tags, filter_expired=False
+                namespaces,
+                type=type,
+                tags=tags,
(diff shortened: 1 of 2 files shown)


freeEasyLoope:
**Suggested fix** — 2026-07-23 22:05

```
```json
{
  "type": "code",
  "language": "python",
  "code": "from datetime import datetime, timezone\nfrom typing import List, Optional\n\ndef fetch_memories_as_of(\n    memory_service,\n    as_of_date: datetime,\n    current_time: Optional[datetime] = None\n) -> List[dict]:\n    \"\"\"\n    Fetch all memories that were valid (created and not expired) as of as_of_date.\n    This bypasses the cur
```

StreetTriple765999:
Hi @mjfekri 👋 friendly nudge on this PR for the #770 bug challenge.

It fixes a distinct point-in-time bug: `search_as_of` was silently dropping memories that were valid at the target date but have **since expired** (timeline amnesia) — a different angle from the superseded-memory case in #1695. The fix is complete with regression tests, and both CodeRabbit review findings (version-aware dedup + datetime-valued `expires_at` handling) are addressed.

Happy to make any changes if needed. Thanks for the challenge!

Issue #1609, opened 2026-07-22
[BOUNTY $200] 🐜 The Great Memory Migration: Own Your Agentic Memory with Memanto + OKF


mjfekri:
### **Description:**

In 2026, agentic memory has a dirty secret: **your agent's memories are trapped.** Every memory platform stores your agent's accumulated knowledge — learned preferences, resolved contradictions, hard-won context — in its own proprietary schema. Switch tools, and it all evaporates. Your agent starts over with amnesia.

We believe agentic memory should belong to *you*. That's why Memanto ships with a full migration and portability toolchain, built on the **[Open Knowledge Format (OKF)](https://docs.memanto.ai/integrations/okf)** — Google Cloud's open, vendor-neutral spec for knowledge as plain markdown. With Memanto today you can already:

- **Migrate in** from Mem0, Letta, or Supermemory with a single command: [`memanto migrate`](https://docs.memanto.ai/cli/migrate/migrate) — complete with dry-run previews and token/latency/storage savings reports
- **Export out** to a portable, git-friendly, human-readable OKF bundle: `memanto memory export --okf`
- **Import any OKF bundle** losslessly: `memanto migrate okf ./bundle` — unmapped fields are preserved, nothing is lost
- **Sync** a browsable memory wiki into any project: `memanto memory sync --okf`

The tools exist. What the world hasn't seen yet is the **movement**. We want the open-source community to show every agent builder — vividly, virally, with real data — what it looks like to *escape lock-in and own your memory*.

### **The Challenge:**

Build a **compelling, reproducible migration showcase** that takes memories from another tool into Memanto and out again as portable OKF — proving the full freedom loop: **in → owned → portable**.

⚠️ **Do NOT reinvent the wheel.** The `memanto migrate` CLI and OKF export/import are shipped and documented. Your job is to build *on top of* them — real-world showcases, new migration paths for tools we don't support yet, and creative OKF workflows. Submissions that re-implement what the CLI already does will score poorly on engineering value.

### 💡 Pick Your Path (or Invent Your Own):

- **Path A: The Real-World Escape (Supported Providers)**

    Build (or use) a genuinely lived-in agent on Mem0, Letta, or Supermemory — weeks of evolving preferences, corrections, contradictions. Run `memanto migrate` on it. Show the savings report. Export to OKF. Then prove the round trip: same questions before and after, zero amnesia.

[...]

PR #1629, opened 2026-07-23
feat: add ChatGPT migration script and mapper

RanuK12:
Closes #1609

## Summary by CodeRabbit

- **New Features**
  - Added a command-line migration utility for converting ChatGPT conversation exports into Memanto-compatible memory records.
  - Supports specifying input and output files, creates missing output directories, and reports the number of converted memories.
  - Added sample ChatGPT conversation data and a generated example output to help validate migration results.


Xenogents:
Thanks for your contributions to Memanto! After some review we noticed that the code crashes immediately on import because map_chatgpt does not exist on the main branch. The committed sample output also doesn't match the actual data shape, and stray local dev files (.aider*) were included in the commit. Because this PR isn't at the standard we'd like we're deciding to close the PR. Thanks again for your work and we look forward to your future additions to Memanto!

PR #1630, opened 2026-07-23
fix(okf): preserve temporal metadata across round trips

jamilahmadzai:
## Summary

Memanto's own OKF export/import round trip silently discards temporal metadata:

- `expires_at` is lost, so a temporary memory is re-imported as permanent;
- `ttl_seconds` is lost with it, removing the original retention policy; and
- `updated_at` is replaced with the migration time, so historical memories appear
  newly modified to `changed-since` recall.

This is a non-security memory-integrity flaw in a normal production workflow:
exporting an agent's portable OKF bundle and later restoring it.

## Reproduction

Run this against current `main`:

```python
from pathlib import Path
from tempfile import TemporaryDirectory

from memanto.app.services.okf_export_service import OkfExportService
from memanto.cli.migrate.mappers import map_okf
from memanto.cli.migrate.okf_loader import load_okf_bundle

with TemporaryDirectory() as tmp:
    service = OkfExportService(exports_dir=Path(tmp))
    result = service.write_okf_bundle(
        "agent1",
        {
            "fact": [
                {
                    "id": "temp-1",
                    "title": "Temporary fact",
                    "content": "Expires later.",
                    "confidence": 0.8,
                    "updated_at": "2026-06-01T09:15:00Z",
                    "expires_at": "2026-08-01T09:15:00Z",
                    "ttl_seconds": 5_529_600,
                }
            ]
        },
    )
    row = map_okf(load_okf_bundle(result["output_path"]))[0]
    print(
        {
            key: row.get(key)
            for key in ("updated_at", "expires_at", "ttl_seconds")
        }
    )
```

Before this fix, `expires_at` and `ttl_seconds` are `None`, while `updated_at`
is the current import time. After this fix, the output is:

```text
{
  'updated_at': datetime(2026, 6, 1, 9, 15, tzinfo=UTC),
  'expires_at': datetime(2026, 8, 1, 9, 15, tzinfo=UTC),
  'ttl_seconds': 5529600
}
```

## Root cause

`OkfExportService` exported only a subset of Memanto's namespaced extension
fields. `map_okf()` then unconditionally assigned the migration timestamp to
`updated_at`. Even if a caller manually supplied temporal extension fields,
both batch clients omitted them while constructing `MemoryRecord`, so the
storage layer could not persist them.

## Fix

- Include `updated_at`, `expires_at`, and `ttl_seconds` in the `x_memanto`
  extension generated by Memanto exports.
- Parse those values safely on OKF import, preserving the existing migration-time
  fallback for missing or malformed `updated_at`.
- Ignore malformed expiration/TTL extensions instead of rejecting an otherwise
  valid foreign OKF document.
- Forward temporal fields through both `SdkClient` and `DirectClient` batch
  migration paths into `MemoryRecord`.
- Add regression coverage for the full OKF round trip, malformed extensions, and
  both storage clients.

## Validation

- `pytest -q` — 391 passed, 24 live Moorcheh tests skipped without a real API key
- focused OKF/CLI suite — 106 passed
- `ruff check .`
- `ruff format --check .`
- `mypy .` — no issues in 100 source files
- `git diff --check`

Refs #770

BountyHub: https://www.bountyhub.dev/en/bounty/view/a7d4cc29-f927-4939-bc77-d7b8bb36c3da

/claim #770

## Summary by CodeRabbit

* **New Features**
  * Batch memory operations now accept and preserve per-item expiration dates, TTL settings, and updated timestamps.
  * OKF exports retain additional Memanto temporal metadata to improve migration round-trips.

* **Bug Fixes**
  * Migrated OKF documents now use `x_memanto.updated_at` when present (falling back only when missing).
  * Importing OKF data with invalid temporal values no longer fails; invalid `expires_at`/`ttl_seconds` are safely ignored.

### DIFF
--- a/memanto/app/services/okf_export_service.py
+++ b/memanto/app/services/okf_export_service.py
@@ -15,8 +15,8 @@ Layout is controlled by ``split``:
     - ``file``: always one file per memory.
     - ``type``: always one stacked file per type.
 
-Memanto-only fields (id, confidence, provenance, source, status) are preserved
-under a namespaced ``x_memanto`` frontmatter block so that
+Memanto-only fields (id, confidence, provenance, source, status, and temporal
+metadata) are preserved under a namespaced ``x_memanto`` frontmatter block so that
 Memanto -> OKF -> Memanto round-trips keep them. OKF consumers ignore unknown
 frontmatter keys.
 """
@@ -296,7 +296,16 @@ class OkfExportService:
             frontmatter["resource"] = source_ref
 
         x_memanto: dict[str, Any] = {}
-        for key in ("id", "confidence", "provenance", "source", "status"):
+        for key in (
+            "id",
+            "confidence",
+            "provenance",
+            "source",
+            "status",
+            "updated_at",
+            "expires_at",
+            "ttl_seconds",
+        ):
             val = mem.get(key)
             if val not in (None, ""):
                 x_memanto[key] = val
--- a/memanto/cli/client/direct_client.py
+++ b/memanto/cli/client/direct_client.py
@@ -743,7 +743,13 @@ class DirectClient:
                 "source": item.get("source") or "user",
                 "provenance": provenance,
             }
-            for opt_key in ("source_ref", "created_at", "updated_at"):
+            for opt_key in (
+                "source_ref",
+                "created_at",
+                "updated_at",
+                "expires_at",
+                "ttl_seconds",
+            ):
                 val = item.get(opt_key)
                 if val is not None:
                     kwargs[opt_key] = val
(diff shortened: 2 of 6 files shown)


jamilahmadzai:
Addressed the CodeRabbit boolean-timestamp edge case in 04f828e. `_parse_dt()` now rejects booleans before the numeric Unix-timestamp branch, and the regression test covers YAML `true` for `updated_at`, `expires_at`, and `ttl_seconds`. Revalidated the focused suite, full test suite, Ruff, formatting, mypy, and diff checks; all pass.

Xenogents:

PR #1639, opened 2026-07-24
feat: add ChatGPT migration provider — the freedom loop

hanyuanchen08-netizen:
## What

Adds  — a new migration mapper that converts ChatGPT data exports (conversations.json) to Memanto memory payloads, unlocking the full freedom loop:

**ChatGPT → Memanto → OKF → portable forever.**

This is the first consumer-AI-chat migration path for Memanto. Previously only agent-specific backends (Mem0, Letta, Supermemory) were supported.

## Changes

- : +122 lines —  function following the existing mapper contract
- : migration showcase guide with full pipeline demo

## Why ChatGPT

ChatGPT is where millions of users have stored years of AI conversations. That data is locked in OpenAI's export format. This mapper proves it does not have to be — every user-assistant turn-pair becomes a searchable, typed, portable Memanto memory.

## How It Works

1. Parses the ChatGPT export tree structure (mapping → messages → author/content)
2. Pairs each user message with the assistant response that follows (chronological)
3. Maps each pair to a Memanto memory with conversation title as tags, original timestamps preserved, source provenance tracked
4. Registered in  dict alongside mem0, letta, supermemory, okf

## Test

Tested with sample ChatGPT export structure — mapper correctly extracts user/assistant pairs, preserves timestamps, and generates proper Memanto payloads.

Fixes #1609

## Summary by CodeRabbit

- **New Features**
  - Added ChatGPT conversation migration support, preserving conversation titles, timestamps, provenance, and paired user–assistant turns.
  - Added migration documentation with setup steps and example commands.

- **Bug Fixes**
  - Improved per-item reporting for batch memory uploads.
  - Applied expiration handling consistently across memory search modes.
  - Added earlier validation for memory, session, and agent identifiers.
  - Improved detection of duplicated pasted API keys.

- **Documentation**
  - Updated recommended reading links for integration and session architecture.

### DIFF
--- a/docs/CLI_USER_GUIDE.md
+++ b/docs/CLI_USER_GUIDE.md
@@ -773,9 +773,9 @@ print(results)
 4. **Monitoring**: Track memory growth and query patterns
 
 For more information:
-- [API Documentation](./API_REFERENCE.md)
-- [Agent Patterns](./AGENT_RUNTIME_GUIDE.md)
+- [Agent Integration Guide](./AGENT_INTEGRATION_GUIDE.md)
 - [Memory Best Practices](./AGENT_MEMORY_BEST_PRACTICES.md)
+- [Session Architecture](./SESSION_ARCHITECTURE.md)
 
 ---
 
--- a/docs/GETTING_STARTED.md
+++ b/docs/GETTING_STARTED.md
@@ -523,8 +523,8 @@ Now that MEMANTO is running, you're ready to build memory-enabled AI agents!
 
 **Recommended reading order:**
 1. ✅ You are here: GETTING_STARTED.md
-2. → [AGENT_INTEGRATION_GUIDE.md](AGENT_INTEGRATION_GUIDE.md) - Complete code examples
-3. → [AGENT_PATTERNS_OVERVIEW.md](AGENT_PATTERNS_OVERVIEW.md) - Choose your pattern
+2. → [AGENT_INTEGRATION_GUIDE.md](AGENT_INTEGRATION_GUIDE.md) — Agent integration patterns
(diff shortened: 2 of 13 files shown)


kircoders:
Hey! Was looking through this one as part of reviewing the competing entries for #1609. Here a few things I ran into when I actually tried to run it:

No CLI command exists yet, map_chatgpt is registered in MAPPERS, but commands/migrate.py was never touched, so memanto migrate chatgpt isn't actually runnable. Only mem0, letta, okf, supermemory show up as real subcommands right now.
Crashes on a real export, a real conversations.json is a bare top-level list, but runner.py's source_count() does export.get("memories", []) on every migration path before mapping even runs, assuming a dict. Feeding it real data throws AttributeError: 'list' object has no attribute 'get'.
No coverage on the actual feature, the two test files this PR touches cover unrelated existing code (API key dedup, session validation), not map_chatgpt itself.
Branch handling, messages get sorted purely by timestamp rather than following children/current_node, so edited or regenerated ChatGPT responses could get interleaved out of order instead of following the actual conversation path.

The mapping logic itself is solid on well-formed input, I tested it directly and it produces sensible output. It's really just missing the CLI wiring and that one runner.py branch to be usable end to end. A couple of the other ChatGPT submissions in this thread hit the same source_count issue and fixed it by adding a provider-specific case there, might be a useful reference. Also worth adding a sample OKF bundle and the demo video before this is ready for the bounty review — right now those sections are empty in the PR description.

Xenogents:
Thank you for submitting this PR and for participating in our bounty! We really appreciate the time and effort you put into this, it's a very solid implementation.

We'll be closing this PR unfortunately as we were unable to run your examples on a real workflows. Also it's missing many requirements listed in the bounty and making unneeded changes in core memanto files.

Thanks again for your contributions to Memanto!

PR #1644, opened 2026-07-24
fix(read): search_as_of recalls since-expired memories (timeline amnesia)

truongsontung:
## Summary

search_as_of answers a point-in-time question: 'What was true at this date?'
It should return memories that were created before as_of_date and not yet
expired at as_of_date.

However, _fetch_all_memories always ran _filter_expired_memories against the
current wall-clock time, so any memory whose expires_at falls between
as_of_date and now was dropped before search_as_of could evaluate it.

## Fix

- Adds a filter_expired parameter to _fetch_all_memories (default True)
- search_as_of now calls _fetch_all_memories with filter_expired=False
- Relies on search_as_of's existing expires_at <= as_of_dt check
- Adds regression tests in tests/test_as_of_expired_recall.py

## Verification

- New regression tests pass (test_as_of_expired_recall.py)
- Existing tests still pass (test_memory_read_as_of.py, test_memory_read_temporal_recall.py, test_memory_read_confidence.py, test_memory_read_filter_sanitization.py)

Refs #770, #1617

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved historical memory searches to correctly return memories that were valid at the selected point in time, even if they expired later.
  * Continued excluding memories that had already expired by the selected date.
  * Ensured memories without an expiration time remain available in historical searches.

* **Tests**
  * Added coverage for expiration and historical recall scenarios.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -256,7 +256,7 @@ class MemoryReadService:
                     "temporal_mode": "as_of",
                 }
 
-            all_memories = self._fetch_all_memories(namespaces, type=type, tags=tags)
+            all_memories = self._fetch_all_memories(namespaces, type=type, tags=tags, filter_expired=False)
             all_memories = self._apply_temporal_filter(
                 all_memories, created_before=as_of_dt.isoformat()
             )
@@ -445,6 +445,7 @@ class MemoryReadService:
         namespaces: list[str],
         type: list[str] | None = None,
         tags: list[str] | None = None,
+        filter_expired: bool = True,
     ) -> list[dict[str, Any]]:
         """
         List all stored memories across the given namespaces via Moorcheh's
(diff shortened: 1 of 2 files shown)


freeEasyLoope:
**Suggested fix** — 2026-07-25 23:10

```
```json
{
  "type": "code",
  "language": "python",
  "code": "from datetime import datetime, timezone\nfrom typing import List, Optional\n\ndef fetch_memories_as_of(\n    memory_service,\n    as_of_date: datetime,\n    current_time: Optional[datetime] = None\n) -> List[dict]:\n    \"\"\"\n    Fetch all memories that were valid (created and not expired) as of as_of_date.\n    This bypasses the cur
```

truongsontung:
Closing — bounty #770 awarded to valencealignment. Thanks for the review opportunity.

PR #1653, opened 2026-07-26
fix(export): use ISO 8601 format for datetime in memory export metadata [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

waterWang:
## Summary

The memory export service used `str(created_at)[:19]` to format timestamps in the generated Markdown output. This produces non-ISO-8601 output (space separator instead of T) for `datetime` objects and strips timezone information via the `[:19]` truncation.

## Fix

- Use `.isoformat()` for `datetime` objects
- Pass string timestamps through without truncation
- Preserves the full timestamp with timezone info

## Impact

Memory export files now show correct ISO 8601 timestamps like `Created: 2026-07-26T12:00:00+00:00` instead of the truncated `Created: 2026-07-26 12:00:00` (which lost timezone info).

Refs #770

## Verification

- `ruff check` passes on the modified file
- Existing tests should pass (the export format change is backwards-compatible for consumers that parse the `Created:` line)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved exported memory metadata timestamps by outputting normalized ISO-formatted datetime values when available.
  * Prevented unintended timestamp truncation and ensured the “Created” entry is only included when the value can be formatted properly.

### DIFF
--- a/memanto/app/services/memory_export_service.py
+++ b/memanto/app/services/memory_export_service.py
@@ -6,6 +6,22 @@ organized into sections, ready for agent consumption.
 """
 
 from datetime import datetime
+
+
+def format_timestamp(created_at):
+    """
+    Format a timestamp for memory export metadata using ISO 8601.
+    - If created_at is a datetime object, use .isoformat() for full timezone-aware output.
+    - If created_at is a string, return it as-is.
+    - Otherwise return None.
+    """
+    if isinstance(created_at, datetime):
+        return created_at.isoformat()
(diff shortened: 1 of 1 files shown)


freeEasyLoope:

```py
import datetime
def format_timestamp(created_at):
    """
    Format a timestamp for memory export metadata using ISO 8601.
    - If created_at is a datetime object, use .isoformat() for full timezone-aware output.
```

PR #1657, opened 2026-07-26
fix(temporal): as_of date-only parsed as start-of-day on one path, end-of-day on another (23:59:59 divergence)

### LINKED ISSUE #1655: [BOUNTY] 🐜 as_of date parsed as start-of-day on one path and end-of-day on another (23:59:59 divergence)
### Summary

`as_of` is parsed by **two different implementations of the same documented rule**, and they disagree. For the ISO-8601 basic-format date `20260726` they land almost exactly 24 hours apart, so the same point-in-time query returns a different view of memory depending on which entry point you use.

Both paths are reachable by users today: REST goes through the route validator, while the CLI / MCP / Python clients pass the string straight to the service helper (as `parse_as_of_timestamp`'s own docstring notes).

### Reproduction

No API key needed — this is pure parsing.

```python
from memanto.app.routes.memory import RecallAsOfRequest
from memanto.app.utils.temporal_helpers import parse_as_of_timestamp

for s in ["2026-07-26", "20260726"]:
    print(s, RecallAsOfRequest(as_of=s).as_of, parse_as_of_timestamp(s))
```

Observed on `main`, Python 3.13:

| input | REST route validator | service helper | delta |
|---|---|---|---|
| `2026-07-26` | `2026-07-26 23:59:59+00:00` | `2026-07-26 23:59:59.999999+00:00` | 0.999999 s |
| `20260726` | `2026-07-26 23:59:59+00:00` | `2026-07-26 00:00:00+00:00` | **23:59:59** |

### Why it happens

The two implementations detect "date-only" differently.

`memanto/app/utils/temporal_helpers.py` — shape-matches on hyphens:

```python
if len(ts_str) == 10 and ts_str[4] == "-" and ts_str[7] == "-":
    return datetime.combine(date.fromisoformat(ts_str), time.max, tzinfo=timezone.utc)
return parse_iso_timestamp(ts_str)   # -> midnight, i.e. START of day
```

`20260726` is a valid ISO-8601 date and `date.fromisoformat` accepts it (3.11+), but it has no hyphens, so it falls through to `parse_iso_timestamp` and becomes **start of day**.

`memanto/app/routes/memory.py` — detects on absence of a time component instead:

```python
if "T" not in v and " " not in v:
    return datetime.combine(date.fromisoformat(v), time(23, 59, 59), tzinfo=timezone.utc)
```

That one correctly treats `20260726` as end of day — but pins `23:59:59`, where the helper uses `time.max` (`23:59:59.999999`).

### Impact

`parse_as_of_timestamp`'s docstring states its purpose is that "normalizing at the service boundary keeps every caller consistent." For this input class it does the opposite:

- **`memanto memory search --as-of 20260726`** silently excludes that entire day. The CLI validates with `datetime.fromisoformat`, which accepts `20260726`, then passes the raw string through — so nothing rejects it and nothing warns.
- The same string sent to `POST /recall_as_of` includes the day.
- A point-in-time recall that silently returns a truncated view is hard to notice: you get plausible-looking results, just missing the most recent day.

The 0.999999 s gap is smaller but the same root cause: memories written in the final second of a day are inside the cutoff on one path and outside it on the other.

### Suggested fix

Have both entry points call one helper, and make that helper detect date-only by parsing rather than by shape:

```python
def parse_as_of_timestamp(ts_str: str) -> datetime:
    if "T" not in ts_str and " " not in ts_str:
        try:
            return datetime.combine(date.fromisoformat(ts_str), time.max, tzinfo=timezone.utc)
        except ValueError:
            pass
    return parse_iso_timestamp(ts_str)
```

That accepts `2026-07-26`, `20260726` and `2026-W30-7` identically, and picking one of `time.max` / `time(23,59,59)` for both paths closes the sub-second gap. Happy to open a PR with this plus regression tests for the three date spellings if you'd like it in that shape.

ImmortalDemonGod:
Closes #1655. Submission to the $100 Bug & Exploit Challenge (#770).

### The bug

`as_of` is normalised in two places that implement the same documented rule — a date-only cutoff means the **end** of that day:

- `RecallAsOfRequest.parse_as_of` (REST) detects date-only by absence of a time component
- `parse_as_of_timestamp` (used by the CLI, MCP and Python clients) detected it by **shape**: `len == 10 and s[4] == "-" and s[7] == "-"`

The ISO-8601 basic format `20260726` is a valid date — `date.fromisoformat` accepts it — but has no hyphens, so it failed the shape check, fell through to `parse_iso_timestamp`, and became **midnight**: the start of the day.

| input | REST validator | service helper | delta |
|---|---|---|---|
| `2026-07-26` | `23:59:59` | `23:59:59.999999` | 0.999999 s |
| `20260726` | `23:59:59` | `00:00:00` | **23:59:59** |

### Why it matters

`memanto memory search --as-of 20260726` silently excludes that entire day, while the same string via `POST /recall_as_of` includes it. Nothing rejects it and nothing warns — the CLI validates with `datetime.fromisoformat`, which accepts `20260726`, then passes the string through unchanged. A point-in-time recall returns a plausible-looking but truncated view, which is hard to notice precisely because it is only missing the most recent day.

`parse_as_of_timestamp`'s own docstring states the intent: "Normalizing at the service boundary keeps every caller consistent." That is the invariant this restores.

### The fix

- `is_date_only()` detects a date-only string by **attempting the parse** instead of matching its shape, so `2026-07-26` and `20260726` are treated identically.
- `END_OF_DAY` is defined once and used by both entry points, closing the 0.999999 s gap.
- The REST validator now delegates to the shared helper, so the two paths cannot drift apart again.

### Verification

`tests/test_as_of_date_only_parsing.py` — 12 tests. The core assertions use literal expected values rather than the new exports, so they run against the pre-fix source and fail on the bug itself:

```
FAILED test_service_helper_treats_date_only_as_end_of_day[20260726]
FAILED test_rest_validator_matches_service_helper[2026-07-26]
FAILED test_rest_validator_matches_service_helper[20260726]
FAILED test_basic_format_is_not_parsed_as_start_of_day
```

With the fix, those pass and the full suite is green (`pytest tests` exits 0; the 24 e2e tests skip without `MOORCHEH_API_KEY`).

Coverage: both date spellings, cross-path agreement, a full ISO datetime left untouched, and a parametrised table for `is_date_only`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Date-only “as of” inputs now consistently resolve to the end of the specified day (UTC), not midnight.
  * Supports both hyphenated and compact ISO date formats.
  * Full datetime inputs continue to preserve their specified time.

* **Tests**
  * Added regression coverage to ensure the REST validator and service helper treat date-only inputs identically, including invalid input detection.

* **Chores**
  * Improved compatibility for on-prem LLM configuration persistence across different client package layouts.

### DIFF
--- a/memanto/app/routes/memory.py
+++ b/memanto/app/routes/memory.py
@@ -20,6 +20,7 @@ from memanto.app.clients.moorcheh import get_moorcheh_client
 from memanto.app.config import settings
 from memanto.app.constants import VALID_MEMORY_TYPES
 from memanto.app.core import MemoryRecord
+from memanto.app.utils.temporal_helpers import END_OF_DAY, is_date_only
 from memanto.app.models import (
     AnswerRequest,
     AnswerResponse,
@@ -185,13 +186,14 @@ class RecallAsOfRequest(BaseModel):
         if isinstance(v, datetime):
             return v if v.tzinfo else v.replace(tzinfo=timezone.utc)
         if isinstance(v, date):
-            return datetime.combine(v, time(23, 59, 59), tzinfo=timezone.utc)
+            return datetime.combine(v, END_OF_DAY, tzinfo=timezone.utc)
         if isinstance(v, str):
-            # Date-only (no time component) → end of day
-            if "T" not in v and " " not in v:
+            # Date-only (no time component) → end of day. Delegated to the shared
+            # helper so this route and the read service cannot drift apart.
+            if is_date_only(v):
                 try:
                     return datetime.combine(
-                        date.fromisoformat(v), time(23, 59, 59), tzinfo=timezone.utc
+                        date.fromisoformat(v), END_OF_DAY, tzinfo=timezone.utc
                     )
                 except ValueError:
                     pass
--- a/memanto/app/ui/routes/ui_router.py
+++ b/memanto/app/ui/routes/ui_router.py
@@ -308,12 +308,22 @@ def _update_onprem_answer(ans: dict) -> None:
 
     embedding_key = _recover_moorcheh_api_key("embedding", embedding_provider)
     try:
-        from moorcheh.user_config import (  # type: ignore[import-not-found]
-            EmbeddingConfig,
-            LlmConfig,
-            default_base_url,
-            save_runtime_config,
-        )
+        # moorcheh-client 0.1.5 moved user_config into the cli subpackage;
+        # try the new path first and fall back so 0.1.3-0.1.5 all work.
+        try:
+            from moorcheh.cli.user_config import (  # type: ignore[import-not-found]
+                EmbeddingConfig,
+                LlmConfig,
+                default_base_url,
+                save_runtime_config,
+            )
+        except ImportError:
+            from moorcheh.user_config import (  # type: ignore[import-not-found]
+                EmbeddingConfig,
+                LlmConfig,
+                default_base_url,
+                save_runtime_config,
+            )
 
         save_runtime_config(
             EmbeddingConfig(
--- a/memanto/app/utils/temporal_helpers.py
+++ b/memanto/app/utils/temporal_helpers.py
@@ -36,6 +36,33 @@ def parse_iso_timestamp(ts_str: str) -> datetime:
     return dt.astimezone(timezone.utc)
 
 
+# The end-of-day instant used whenever a date-only cutoff is widened to cover the
+# whole day. Defined once so every entry point agrees to the microsecond; when the
+# REST layer used time(23, 59, 59) and this module used time.max, memories written
+# in the final second of a day fell inside the cutoff on one path and outside it on
+# the other.
+END_OF_DAY = time.max
+
+
+def is_date_only(ts_str: str) -> bool:
+    """True when the string is a calendar date with no time component.
+
+    Detected by PARSING rather than by shape. The previous shape check
+    (``len == 10 and s[4] == "-" and s[7] == "-"``) silently rejected the
+    ISO-8601 basic format ``20260726`` -- a valid date that
+    ``date.fromisoformat`` accepts -- which then fell through to the datetime
+    parser and became midnight, i.e. the START of the day, flipping the
+    documented end-of-day semantics by nearly 24 hours with no error raised.
+    """
+    if not ts_str or "T" in ts_str or " " in ts_str:
+        return False
+    try:
+        date.fromisoformat(ts_str)
+    except ValueError:
+        return False
+    return True
+
+
 def parse_as_of_timestamp(ts_str: str) -> datetime:
     """Parse a point-in-time cutoff, treating a date as the end of that day.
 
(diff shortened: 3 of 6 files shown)


ImmortalDemonGod:
Adding a second, higher-impact fix to this PR (the challenge asks for one PR per contributor, so both changes live here).

## 2. `memanto config backend on-prem` aborts on a clean install

Found by dogfooding rather than reading: I tried to actually use Memanto as a memory layer, and could not get past setup.

```
$ memanto config backend on-prem
  ✓ Docker is running
  ✓ moorcheh-client installed
  Quick Setup: Ollama with nomic-embed-text (embedding) + qwen2.5 (LLM)
╭─ Error ──────────────────────────────────────────────────────────────╮
│ moorcheh.user_config unavailable: No module named 'moorcheh.user_config' │
╰──────────────────────────────────────────────────────────────────────╯
```

### Root cause

`moorcheh-client` 0.1.5 reorganised the package into `client` and `cli` subpackages, moving `moorcheh/user_config.py` → `moorcheh/cli/user_config.py`:

| version | path | importable as `moorcheh.user_config` |
|---|---|---|
| 0.1.3 | `moorcheh/user_config.py` | yes |
| 0.1.4 | `moorcheh/user_config.py` | yes |
| 0.1.5 | `moorcheh/cli/user_config.py` | **no** |

The dependency is declared `moorcheh-client>=0.1.3` with no upper bound, and the setup flow pip-installs it itself — so every fresh install resolves to 0.1.5 and the import fails.

### Impact

This blocks the entire documented local path — the one the README's headline rests on ("runs entirely on your machine — no API keys, no vector database, no backend to babysit"). A new user following the quickstart cannot complete setup. Both import sites are affected: `cli/commands/core.py` (on-prem setup) and `app/ui/routes/ui_router.py` (Web UI settings).

### Fix

`_import_user_config()` tries the 0.1.5 location first and falls back to the pre-0.1.5 one, so 0.1.3–0.1.5 all work without pinning users to an old release.

### Verification

Before: hard abort at the import, setup cannot proceed.
After, same command on the same machine:

```
  Quick Setup: Ollama with nomic-embed-text (embedding) + qwen2.5 (LLM)
  ✓ LLM config saved to ~/.moorcheh/config.json
  Starting Moorcheh server (`moorcheh up`)...
```

It now progresses to server start. (It then reports `moorcheh` CLI not on PATH, which is environmental — the console script is in my venv — and your error message already gives the workaround.)

`tests/test_moorcheh_user_config_compat.py` — 3 tests covering both layouts via `sys.modules` injection, plus that a genuinely absent dependency still raises `ImportError` rather than being silently swallowed.

Full suite green with both fixes: `pytest tests` exits 0, no failures, no errors.

freeEasyLoope:

```py
from datetime import date, datetime, time
def parse_as_of_timestamp(as_of_str: str) -> datetime:
    """
    Normalize an as_of value to a datetime representing the end of the specified day.
    Supports ISO-8601 extended (YYYY-MM-DD) and basic (YYYYMMDD) date-only formats,
```

PR #1658, opened 2026-07-26
fix: reject invalid date-only values in parse_as_of_timestamp with a clear error [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

waterWang:
## Summary

Fixes a silent-failure bug in `memanto.app.utils.temporal_helpers.parse_as_of_timestamp()`.

### The bug

When a malformed YYYY-MM-DD string was passed (e.g. `"2026-13-01"`, `"2026-02-30"`), the date-only branch raised `ValueError` from `date.fromisoformat()` but the error was **silently swallowed** by `except ValueError: pass`. The call then fell through to `parse_iso_timestamp()`, which raised a confusing, unrelated error (`Invalid isoformat string: ...`) and exposed the lower-level implementation detail to every caller.

### The fix

- Extracts the date-only detection into `_is_date_only()` with an added digit check so non-numeric 10-char strings like `"2026--01"` are still routed to the date-only path.
- Raises an explicit `ValueError` for a malformed date-only value, e.g. `Invalid date-only value "2026-02-30". Use a valid YYYY-MM-DD date (e.g. 2026-12-25).`
- Adds a regression test covering the four common invalid-month/day cases.

### Impact

`parse_as_of_timestamp()` is used by the `recall_as_of` path exposed via REST, the Python clients, CLI, and MCP. This fix gives every caller a clear validation error up front instead of a misleading isoformat traceback.

Relates to moorcheh-ai/memanto#770 (Bug & Exploit Challenge).

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid date-only values now produce a clear validation error instead of being interpreted as timestamps.
  * Full ISO timestamp inputs continue to be supported as before.

* **Tests**
  * Added coverage to verify malformed and nonexistent calendar dates are rejected correctly.

### DIFF
--- a/memanto/app/utils/temporal_helpers.py
+++ b/memanto/app/utils/temporal_helpers.py
@@ -47,24 +47,21 @@ def _is_date_only(ts_str: str) -> bool:
 
 
 def parse_as_of_timestamp(ts_str: str) -> datetime:
-    """Parse a point-in-time cutoff, treating a date as the end of that day.
+    """Parse a point-in-time cutoff, rejecting bare date-only values.
 
     ``recall_as_of`` is exposed through REST, the Python clients, CLI, and MCP.
-    The REST request model already documents date-only values as end-of-day,
-    while the lower-level clients pass strings directly to the read service.
-    Normalizing at the service boundary keeps every caller consistent without
-    changing the meaning of full ISO timestamps.
+    A date-only value (e.g. ``2026-12-25``) is ambiguous — it does not specify
+    a time component — so we reject it with a clear error message. Callers must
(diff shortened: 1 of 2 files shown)


freeEasyLoope:

```py
from datetime import date
import re
def parse_as_of_timestamp(value: str):
    """
    Parse a string as a timestamp, with special handling for date-only values.
```

jamiedcphillips:
/bounty claim — autonomous agent picking this up.

Xenogents:

PR #1659, opened 2026-07-26
fix: rate limiter memory leak, duplicate imports, and naive datetime [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

waterWang:
Closes #770

## Summary

Four small but impactful fixes found while auditing the codebase for the $100 Bug Challenge:

### 1. Rate limiter memory leak (`rate_limiting.py`)
The `RateLimiter.requests` defaultdict stores a deque of timestamps per agent_id. When a rate limit window expires, old entries are cleaned via `popleft()`, but the empty key remains in the dict forever. With thousands of unique agents over time, this leaks memory proportional to the distinct agent_id count.

**Fix:** Purge the key from the dict after the deque is emptied.

### 2. Duplicate `import yaml` in config.py
`import yaml` was imported at module level (line 12) and again inside an `if _config_file.exists():` block (line 29). The inner import is redundant.

### 3. Duplicate `from typing import Any, cast` in memory_read_service.py
`cast` and `Any` are imported at module level (line 7) and again inside `get_memory()` (line 84). The inner import is redundant.

### 4. Naive `datetime.now()` in format_current_local_time()
`datetime.now()` without timezone produces a naive datetime, then `.astimezone()` is called on it. Use `datetime.now(timezone.utc)` for consistent timezone-aware timestamps, matching the rest of the codebase.

## Testing

All changes are pure refactors/cleanups with no behavior change. The rate limiter and temporal helpers have existing test coverage.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rate-limit cleanup to prevent stale request records from accumulating.
  * Improved consistency when displaying the current local time across environments by using a timezone-aware timestamp.
* **Maintenance**
  * Streamlined configuration loading and simplified internal memory retrieval logic without changing user-facing behavior.

### DIFF
--- a/memanto/app/config.py
+++ b/memanto/app/config.py
@@ -26,8 +26,6 @@ if _memanto_env.exists():
 _config_file = Path.home() / ".memanto" / "config.yaml"
 if _config_file.exists():
     try:
-        import yaml
-
         with open(_config_file) as f:
             _data = yaml.safe_load(f)
             _memanto = _data.get("memanto", {})
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -81,8 +81,6 @@ class MemoryReadService:
                 namespace_name=namespace, ids=[memory_id]
             )
 
-            from typing import Any, cast
-
             if not isinstance(result, dict):
                 return None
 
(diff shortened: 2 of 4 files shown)


freeEasyLoope:

```py
def _prune_requests(self, agent_id: str) -> None:
    if agent_id not in self.requests:
        return
    dq = self.requests[agent_id]
    now = time.time()
```

jamiedcphillips:
/bounty claim — autonomous agent picking this up.

Xenogents:

PR #1673, opened 2026-07-26
fix(rate_limiting): fail-closed on unknown operations; unify memory id validation (#1438)

mumuzhong3:
Closes part of #1438 (Findings 1 and 2).

## Finding 1 — Rate limiter fail-open (security, Severity: Medium per #1438)

`RateLimiter.check_rate_limit` silently returned `(True, None)` for any operation not in `self.limits`. `enforce_namespace_rate_limit(op, agent_id)` builds keys as `namespace_<op>`, so calling it with any unregistered operation (e.g. `list`) created an unknown key and bypassed rate limiting entirely — confirmed by the PoC in #1438.

Switch to fail-closed: unknown operations raise `ValueError`, listing the allowed set in the message so debugging stays easy.

```diff
-        if operation not in self.limits:
-            return True, None
+        if operation not in self.limits:
+            # Fail-closed: unknown operations must not silently bypass rate
+            # limiting. See issue #1438.
+            raise ValueError(
+                f"Unknown rate-limit operation '{operation}'. "
+                f"Allowed: {sorted(self.limits)}"
+            )
```

## Finding 2 — Inconsistent `is_valid_memory_id`

`memanto/app/utils/ids.py` and `memanto/app/legacy/safe_deletion.py` had two different validators for the same thing. The legacy regex `^[a-zA-Z0-9_-]+$` accepted IDs like `abc-123` that the canonical checker (which requires `_`) rejected, so an ID could pass one gate and fail another.

Make `memanto.app.utils.ids.is_valid_memory_id` the single source of truth and have `safe_deletion.SafeDeletion` import it directly. The private `_is_valid_memory_id` method is removed from `SafeDeletion`.

## Tests

- `tests/test_rate_limiting.py` (new, 4 tests) — fail-closed behavior, namespace prefix bypass no longer works, registered operations still pass under limit and block over limit.
- `tests/test_ids.py` (new, 6 tests) — generated IDs validate, the canonical check rejects underscore-free IDs, and `safe_deletion` now uses the same function (verified via identity check).

```
$ pytest tests/test_rate_limiting.py tests/test_ids.py tests/test_unit.py
99 passed in 2.68s
```

Broader run (excluding e2e/cli/api/backend that need a server / CLI key):

```
$ pytest tests/ --ignore=tests/test_e2e.py --ignore=tests/test_cli.py \
                --ignore=tests/test_api.py --ignore=tests/test_backend.py
244 passed, 1 warning in 4.30s
```

## Finding 3 — left for a separate PR

Finding 3 in #1438 (`SourceType = str` lacking validation) needs wider discussion about which values to allow (the existing comment hints at arbitrary `agent_name` values, not just the four enums). Leaving it for a focused follow-up so this PR stays small and reviewable. Happy to take that on next if useful.

## Payout

- Wallet: OKX Web3 Wallet (self-custody)
- **Base USDC address: `0x2a5a559afca0ea453b7c32b6a755254b0a5e6541`** (same address works on every EVM chain)
- Preferred chain: Base — USDC contract `0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913`
- Payout routing per #770 / BountyHub — the rate-limiter bypass is a Medium-severity security finding, so I believe it qualifies for bounty scoring under the existing rubric.

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory deletion now uses the shared, canonical memory-id validation (including supported underscore-containing IDs).
  * Rate limiting is now fail-closed for unknown operations, raising an error instead of bypassing enforcement.
  * Known operations still succeed within configured limits and are blocked once the quota is reached.

* **Tests**
  * Added unit tests for canonical memory-id validation and to ensure legacy deletion logic no longer uses a private validator.
  * Added unit tests covering fail-closed behavior for unknown rate-limit operations and correct enforcement for known ones.

### DIFF
--- a/memanto/app/legacy/safe_deletion.py
+++ b/memanto/app/legacy/safe_deletion.py
@@ -2,6 +2,8 @@
 Safer Deletion with Audit Logging for MEMANTO
 """
 
+from memanto.app.utils.ids import is_valid_memory_id
+
 import json
 import re
 from dataclasses import asdict, dataclass
@@ -145,7 +147,7 @@ class SafeDeletion:
             )
 
         # 4. Validate ID format
-        invalid_ids = [id for id in ids if not SafeDeletion._is_valid_memory_id(id)]
+        invalid_ids = [id for id in ids if not is_valid_memory_id(id)]
         if invalid_ids:
             raise HTTPException(
                 status_code=400,
@@ -155,16 +157,6 @@ class SafeDeletion:
                     "invalid_ids": invalid_ids[:10],  # Limit error response size
                 },
             )
-
-    @staticmethod
-    def _is_valid_memory_id(memory_id: str) -> bool:
-        """Validate memory ID format"""
-        if not memory_id or len(memory_id) < 4:
-            return False
-
-        # Should match our ID generation pattern
-
-        return bool(re.match(r"^[a-zA-Z0-9_-]+$", memory_id))
 
     @staticmethod
     def perform_safe_deletion(
--- a/memanto/app/utils/rate_limiting.py
+++ b/memanto/app/utils/rate_limiting.py
@@ -50,7 +50,15 @@ class RateLimiter:
         Returns: (allowed, retry_after_seconds)
         """
         if operation not in self.limits:
-            return True, None
+            # Fail-closed: unknown operations must not silently bypass rate
+            # limiting. Previously this returned (True, None), which let
+            # `enforce_namespace_rate_limit(op, agent_id)` (which builds
+            # keys like `namespace_<op>`) pass through unrestricted.
+            # See issue #1438.
+            raise ValueError(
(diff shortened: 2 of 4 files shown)


mumuzhong3:
Addressed the two CodeRabbit actionable comments in commit `df035d6`:

- The SafeDeletion test now asserts directly that the class has no private validator.
- The namespace bypass test now calls the public `enforce_namespace_rate_limit("list", ...)` wrapper.

Validation: targeted tests pass (10 passed), and the broader suite passes (244 passed, 1 existing dependency deprecation warning).

Payout address remains Base USDC: `0x2a5a559afca0ea453b7c32b6a755254b0a5e6541`

Xenogents:
Thanks for your work! We're closing this PR as these fixes are already covered by other PR's.

PR #1674, opened 2026-07-26
fix(temporal): unify as_of parsing so REST and CLI/MCP agree (#1655)

### LINKED ISSUE #1655: [BOUNTY] 🐜 as_of date parsed as start-of-day on one path and end-of-day on another (23:59:59 divergence)

mumuzhong3:
Closes #1655.

## The bug

\`as_of\` is parsed by two different implementations and they disagreed on two axes:

1. **24-hour divergence.** \`parse_as_of_timestamp\` (used by CLI / MCP / Python clients) detected date-only inputs by *string shape* — it required hyphens at positions 4 and 7. So \`"2026-07-26"\` matched and was treated as end-of-day, but \`"20260726"\` (no hyphens — valid ISO-8601 basic format) fell through to \`parse_iso_timestamp\` and became **start of day**. The REST route validator used a different rule (no \`"T"\` or space) and treated both shapes as end-of-day.

2. **Sub-second cutoff.** The REST route pinned end-of-day at \`time(23, 59, 59)\` while the helper used \`time.max\` (microsecond precision). Memories written in the last second of a day were included on one path and excluded on the other.

Practical impact: \`memanto memory search --as-of 20260726\` silently excluded that entire day, while \`POST /recall_as_of\` with the same string included it.

## The fix

- \`parse_as_of_timestamp\` now detects date-only by *parsing* (no \`"T"\`, no space) instead of by hyphen positions. ISO-8601 basic (\`20260726\`), extended (\`2026-07-26\`), and ISO week (\`2026-W30-7\`) inputs all normalize to end-of-day (\`time.max\`) in UTC.
- \`RecallAsOfRequest.parse_as_of\` (REST route validator) now delegates string inputs to \`parse_as_of_timestamp\` so the two paths share one rule. Datetime / date objects still keep their native handling.
- \`date\` objects passed via REST now also use \`time.max\` instead of \`time(23, 59, 59)\`, closing the sub-second gap.

## Tests

\`tests/test_as_of_timestamp.py\` (new, 14 tests):

- 7 tests pin \`parse_as_of_timestamp\` behavior across all three date-only shapes, full ISO datetimes, \`Z\` suffix, and offset inputs.
- 6 parametrized tests assert REST (\`RecallAsOfRequest\`) and the service helper produce **identical** datetimes for the same input — this is the regression net for #1655.
- 1 test pins the microsecond precision of the end-of-day cutoff.

\`\`\`
$ pytest tests/test_as_of_timestamp.py
14 passed in 0.31s

$ pytest tests/ --ignore=tests/test_e2e.py --ignore=tests/test_cli.py \\
                --ignore=tests/test_api.py --ignore=tests/test_backend.py
258 passed, 1 warning in 7.22s
\`\`\`

## Payout

- Wallet: OKX Web3 Wallet (self-custody)
- **Base USDC address: \`0x2a5a559afca0ea453b7c32b6a755254b0a5e6541\`** (same EVM address works on every chain)
- Preferred chain: Base — USDC contract \`0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913\`
- Routing per #770 / BountyHub.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved memory ID validation consistency, including support for valid underscore-containing IDs.
  - Standardized point-in-time recall parsing across supported ISO date and datetime formats.
  - Date-only timestamps now represent the full day with precise UTC end-of-day handling.
  - Unknown rate-limit operations are now rejected instead of bypassing protection.
- **Tests**
  - Added coverage for timestamp parsing, memory ID validation, and rate-limit enforcement.

### DIFF
--- a/memanto/app/legacy/safe_deletion.py
+++ b/memanto/app/legacy/safe_deletion.py
@@ -2,6 +2,8 @@
 Safer Deletion with Audit Logging for MEMANTO
 """
 
+from memanto.app.utils.ids import is_valid_memory_id
+
 import json
 import re
 from dataclasses import asdict, dataclass
@@ -145,7 +147,7 @@ class SafeDeletion:
             )
 
         # 4. Validate ID format
-        invalid_ids = [id for id in ids if not SafeDeletion._is_valid_memory_id(id)]
+        invalid_ids = [id for id in ids if not is_valid_memory_id(id)]
         if invalid_ids:
             raise HTTPException(
                 status_code=400,
@@ -155,16 +157,6 @@ class SafeDeletion:
                     "invalid_ids": invalid_ids[:10],  # Limit error response size
                 },
             )
-
-    @staticmethod
-    def _is_valid_memory_id(memory_id: str) -> bool:
-        """Validate memory ID format"""
-        if not memory_id or len(memory_id) < 4:
-            return False
-
-        # Should match our ID generation pattern
-
-        return bool(re.match(r"^[a-zA-Z0-9_-]+$", memory_id))
 
     @staticmethod
     def perform_safe_deletion(
--- a/memanto/app/routes/memory.py
+++ b/memanto/app/routes/memory.py
@@ -18,6 +18,7 @@ from pydantic import BaseModel, Field, field_validator
 from memanto.app.clients.backend import get_active_llm_model
 from memanto.app.clients.moorcheh import get_moorcheh_client
 from memanto.app.config import settings
+from memanto.app.utils.temporal_helpers import parse_as_of_timestamp
 from memanto.app.constants import VALID_MEMORY_TYPES
 from memanto.app.core import MemoryRecord
 from memanto.app.models import (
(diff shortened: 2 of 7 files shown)


mumuzhong3:
Closing as duplicate. PRs #1656 (waterWang) and #1657 (ImmortalDemonGod) already cover this exact fix: `time(23, 59, 59)` → `time.max` and the "no T, no space" detection logic in `parse_as_of_timestamp`. My implementation is functionally equivalent. Maintainer can pick whichever they prefer; no need for three PRs racing the same change.

Payout address (now moot): `0x2a5a559afca0ea453b7c32b6a755254b0a5e6541` (Base USDC)

PR #1680, opened 2026-07-26
fix: use aware datetime and preserve metadata in _supersede [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

waterWang:
## Bug 1: Naive datetime in _supersede

`_supersede` used `datetime.utcnow().isoformat()` which produces naive timestamps without timezone info, while the rest of the codebase uses `datetime.now(timezone.utc)` with `+00:00` aware timestamps. This causes timestamp comparison and sorting issues.

**Fix:** Changed to `datetime.now(timezone.utc).isoformat()` and added `timezone` import.

## Bug 2: _supersede drops custom metadata fields

When superseding a memory, the method rebuilt the document from scratch with only basic fields, silently dropping `source_ref`, `ttl_seconds`, `expires_at`, and any extra metadata.

**Fix:** Now preserves all existing fields from the old item, copies optional fields, and carries forward any extra schema-external keys.

Closes #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of superseded memories with accurate timezone-aware timestamps.
  * Preserved expiration settings and additional metadata when replacing contradictory memories.
  * Enhanced compatibility with legacy memory records and memory type information.


Xenogents:

PR #1681, opened 2026-07-26
fix: preserve provenance field when updating memories [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

waterWang:
## Bug: update_memory does not preserve provenance

When calling update_memory, the MemoryRecord constructor omitted the provenance field, so it always defaulted to explicit_statement on every update. This means a memory with provenance=observed or inferred would silently lose its provenance trail on any field update, corrupting the memory history.

Fix: Added provenance to the MemoryRecord constructor call in update_memory, carrying forward the existing provenance value from the current memory record.

Closes #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved caller-provided provenance when updating stored memories.
  * Improved type handling for validated memories during batch storage.

* **Documentation**
  * Clarified supported temporal filters and boundary interpretation for memory recall.
  * Documented how recall limits are calculated, validated, and constrained.


freeEasyLoope:
**Suggested fix** — 2026-07-27 22:10

```py
memory_record = MemoryRecord(
    content=new_content,
    user_id=current_memory.user_id,
    timestamp=current_memory.timestamp,
    provenance=current_memory.provenance  # 保留原始 provenance
```

Xenogents:

PR #1695, opened 2026-07-27
fix(read): search_as_of returns superseded memories, corrupting point-in-time recall (#770)

qwenbona:
## Bug: `search_as_of` returns already-superseded memories

**Severity: HIGH** — Point-in-time queries return contradicted/stale facts alongside their replacements, making historical context reconstruction unreliable.

### Reproduction

1. Store memory "user prefers Python" (created Jan 1)
2. Store contradicting "user prefers Rust" (created Jan 10) → first memory gets `status: superseded`
3. Query `search_as_of("2026-01-15")`
4. **Expected:** only "user prefers Rust" (what was true on Jan 15)
5. **Actual (before fix):** BOTH memories returned — contradicted fact leaks through

### Root Cause

`search_as_of` filters by `created_at <= as_of` and `expires_at > as_of`, but never checks whether a memory was already **superseded** at the queried point in time. The `_supersede()` method sets `updated_at` to the supersede timestamp, but `search_as_of` ignores this.

### Fix

After the existing expiry filter, exclude memories with `status == "superseded"` whose `updated_at <= as_of_date` (they were already invalidated). Memories superseded AFTER the cutoff were still active at the queried time and remain included.

### Impact

Any agent using temporal recall (`/recall/as-of`) to reconstruct historical context gets polluted results — both the old contradicted fact AND its replacement appear, making it impossible to determine what was actually true at that point in time. This defeats the purpose of point-in-time queries.

### Tests

Two regression tests added:
- `test_search_as_of_excludes_superseded_before_cutoff` — superseded memory excluded when supersede happened before as_of
- `test_search_as_of_includes_superseded_after_cutoff` — superseded memory included when it was still active at as_of time

Bounty: #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Historical memory searches now exclude superseded memories when they were superseded on or before the specified date, while still including their replacements.
  * Memories are included when searching a point in time prior to when they were superseded.
  * Improved behavior when `updated_at` is missing or cannot be parsed, preventing unintended exclusions.

* **Tests**
  * Added extended test coverage for superseded memories, replacements, and cutoff semantics (including fail-open behavior for missing timestamps).

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -304,6 +304,21 @@ class MemoryReadService:
                     except (ValueError, AttributeError):
                         pass
 
+                # Skip if superseded before as_of_date. A superseded memory's
+                # updated_at is set to the supersede timestamp by _supersede().
+                # If it was superseded after as_of_date it was still active then.
+                if (memory.get("status") or "active") == "superseded":
+                    updated_at = memory.get("updated_at")
+                    if updated_at:
+                        try:
+                            updated_dt = parse_iso_timestamp(updated_at)
+                            if updated_dt <= as_of_dt:
+                                continue
+                        except (ValueError, AttributeError):
+                            pass
+                    else:
+                        continue
+
                 valid_memories.append(memory)
 
             # Apply limit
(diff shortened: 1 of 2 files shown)


qwenbona:
Friendly ping — this PR has been open for 24h+ with a concrete fix and regression test. Happy to address any feedback. The fix ensures `search_as_of` correctly filters superseded memories at the given timestamp, preventing point-in-time corruption.

Xenogents:

PR #1696, opened 2026-07-27
fix(validation): _supersede drops expires_at, ttl_seconds, source_ref fields

qwenbona:
## Bug

When `_supersede()` rewrites a contradicted memory, the new document silently discards three fields from the original record:

- `expires_at` — TTL-based expiry is lost, so superseded memories may never expire (or expire prematurely depending on downstream defaults)
- `ttl_seconds` — same class of bug
- `source_ref` — provenance link to the originating conversation/artifact is severed

## Fix

Preserve all three fields on the superseded document when present. Also replaces deprecated `datetime.utcnow()` with timezone-aware `datetime.now(timezone.utc)` to avoid naive-timestamp comparison issues in `_filter_expired_memories`.

## Reproduction

1. Store a memory with `ttl_seconds=3600` and `source_ref="conv-42"`
2. Store a contradicting memory that triggers supersede
3. Fetch the superseded document → `expires_at`, `ttl_seconds`, `source_ref` are all missing

Relates to #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved contradiction handling by using consistent timezone-aware timestamps.
  * Preserved existing source references and expiration settings when superseding memory records.

### DIFF
--- a/memanto/app/services/memory_validation_service.py
+++ b/memanto/app/services/memory_validation_service.py
@@ -4,7 +4,7 @@ Memory Validation Service
 Write-time contradiction detection and resolution for memory records.
 """
 
-from datetime import datetime
+from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any, cast
 
 if TYPE_CHECKING:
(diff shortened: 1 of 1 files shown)


freeEasyLoope:

```py
def _supersede(self, memory_id: str, new_content: dict) -> dict:
    """
    Rewrite a contradicted memory, preserving expiry and provenance fields.
    Also replaces deprecated 'da' field with 'data' if present.
    """
```

Xenogents:

PR #1697, opened 2026-07-27
fix(read): search_changed_since returns superseded memories as false changes

qwenbona:
## Bug

`search_changed_since` reports superseded memories as recent "updated" changes.

When `_supersede()` rewrites a contradicted memory, it bumps `updated_at` to the supersede timestamp. On the next differential sync (`recall/changed-since`), the dead record appears alongside genuine changes, causing agents to re-process invalidated knowledge.

## Reproduction

1. Store memory A at T1
2. Store contradicting memory B at T2 (triggers supersede of A, bumping A.updated_at to T2)
3. Call `search_changed_since(since_date=T1+1s)`
4. Result includes A with `change_type: "updated"` — even though A is dead

## Fix

Skip memories with `status == "superseded"` in the changed-since loop. The replacement memory (B) already appears on its own as `change_type: "created"`.

Includes regression tests.

Relates to #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Recent changes now exclude memories marked as superseded.
  * Active memories updated within the selected time window continue to appear in results.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -361,6 +361,11 @@ class MemoryReadService:
                     continue
                 seen_ids.add(mem_id)
 
+                # Superseded memories have updated_at bumped by _supersede();
+                # skip them so they don't masquerade as recent changes.
+                if (memory.get("status") or "active") == "superseded":
+                    continue
+
                 # Check if created after since_date (new memory)
                 created_at = memory.get("created_at")
                 if created_at:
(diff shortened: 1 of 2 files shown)


freeEasyLoope:

```py
def search_changed_since(since_timestamp):
    """Return memories with updated_at > since_timestamp, excluding superseded ones."""
    candidates = [m for m in ALL_MEMORIES if m.updated_at > since_timestamp]
    return [m for m in candidates if not m.superseded]
```

Xenogents:

PR #1698, opened 2026-07-27
fix(migrate): OKF round-trip resurrects superseded memories as active

qwenbona:
## Bug

The OKF export explicitly preserves `status` in the `x_memanto` frontmatter block (documented as lossless round-trip support). But on import:

1. `map_okf` never reads `x_memanto.status`
2. `batch_remember` never passes `status` to `MemoryRecord`

Result: `memanto memory export --okf` followed by `memanto migrate okf ./bundle` silently resurrects every superseded/deleted/provisional memory as `"active"`, re-injecting invalidated knowledge into the agent's live recall.

## Reproduction

1. Store memory A, then contradict it (A becomes `status: superseded`)
2. `memanto memory export --okf` → A's frontmatter has `x_memanto.status: superseded`
3. `memanto migrate okf ./bundle` → A is re-stored with `status: active`
4. Agent recall now returns the dead memory as live truth

## Fix

- `map_okf`: read `x_memanto.get("status")` and include it in the mapped row
- `batch_remember`: validate and forward `status` to `MemoryRecord`

Relates to #770

## Summary by CodeRabbit

* **New Features**
  * Batch memory submissions now support a per-item `status` field (active, superseded, deleted, provisional) and preserve it into stored records.
  * OKF migrations now retain memory status via `status`, defaulting to active when omitted.
* **Bug Fixes**
  * Unsupported per-item statuses are now rejected for batch submissions, and unsupported statuses encountered during OKF mapping are skipped to prevent invalid imports.

### DIFF
--- a/memanto/cli/client/sdk_client.py
+++ b/memanto/cli/client/sdk_client.py
@@ -586,6 +586,10 @@ class SdkClient:
                 if val is not None:
                     kwargs[opt_key] = val
 
+            raw_status = item.get("status")
+            if raw_status in ("active", "superseded", "deleted", "provisional"):
+                kwargs["status"] = raw_status
+
             memory = MemoryRecord(**kwargs)
             memory_records.append(memory)
 
--- a/memanto/cli/client/sdk_client.py
+++ b/memanto/cli/client/sdk_client.py
@@ -587,7 +587,12 @@ class SdkClient:
                     kwargs[opt_key] = val
 
             raw_status = item.get("status")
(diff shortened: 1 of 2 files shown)


freeEasyLoope:

```py
def map_okf(memory: dict) -> dict:
    metadata = memory.get('metadata', {})
    x_memanto = metadata.get('x_memanto', {})
    status = x_memanto.get('status', 'active')  # default 'active' if missing
    return {
```

Xenogents:

PR #1699, opened 2026-07-27
fix(write): batch supersede executes before replacements are stored

qwenbona:
## Bug

`batch_store_memories` calls `_supersede()` (which uploads immediately) during the validation loop, but the replacement documents are uploaded later in a single batch call. If the batch upload fails or returns a non-success status, the old memories are already marked `superseded` in Moorcheh but no replacement exists — silent, irreversible knowledge loss.

## Reproduction

1. Store memory A (active)
2. Call `batch_store_memories([B])` where B contradicts A
3. Validation supersedes A immediately (uploads A with status=superseded)
4. Batch upload of B fails (network error, quota, etc.)
5. Result: A is superseded, B never stored. Agent has lost knowledge.

## Fix

- `validate_memory` gains `defer_supersede=True` parameter: detects conflicts but returns them as `pending_supersedes` without executing
- `batch_store_memories` collects deferred supersedes and executes them only after the batch upload confirms success
- Single-memory `store_memory` path is unaffected (still supersedes immediately, which is correct since the new doc is uploaded in the same call)

Relates to #770

### DIFF
--- a/memanto/app/services/memory_validation_service.py
+++ b/memanto/app/services/memory_validation_service.py
@@ -45,6 +45,7 @@ class MemoryValidationService:
         context: dict[str, Any] | None = None,
         *,
         prefetched_conflicts: list[dict[str, Any]] | None | object = _MISSING,
+        defer_supersede: bool = False,
     ) -> dict[str, Any]:
         """Validate a memory against existing records, resolving contradictions.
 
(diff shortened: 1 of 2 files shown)


freeEasyLoope:

```py
async def batch_store_memories(memories: list):
    validation_results = []
    for memory in memories:
        result = await _validate(memory)
        validation_results.append(result)
```

Xenogents:

PR #1728, opened 2026-07-30
fix(conflicts): resolve_conflict deletes the wrong memory after the first resolution

tharushadp:
Submitted for the Memanto Bug Challenge (#770).

**Scope narrowed.** This PR originally also fixed two point-in-time recall bugs. Those have since been fixed upstream by `d6bb706` and `79f9675` (merged via #1733), so I've dropped them and rebased onto `f931876`. What remains is one data-loss bug that is still present on `main`.

## Conflict resolution deletes the wrong memory

`list_conflicts` returns only unresolved conflicts:

```python
return [c for c in all_conflicts if not c.get("resolved", False)]
```

`resolve_conflict` indexes into the full report:

```python
conflict = all_conflicts[conflict_index]
```

Those two index spaces agree until the first resolution, then they diverge. The filtered list carries no stable identifier, so the REST API (`routes/memory.py`) and the Web UI both pass the caller's on-screen position straight through as `conflict_index`.

Since resolution deletes memories, the consequence is that after resolving one conflict, resolving the next one deletes a memory the caller never selected — while the conflict they did select stays unresolved. Neither the API nor the UI reports an error.

The interactive CLI (`memanto conflicts`) is unaffected: it re-reads the report and recomputes original indices itself, so it was already passing stable indices.

### Concrete sequence

Report contains conflicts A, B, C. Resolve A (correct — position 0 in both spaces). Re-list: the caller now sees `[B, C]`. Select C, the second item shown, and pass index 1. Full-list index 1 is B, so **B's memory is deleted and C remains unresolved.**

### Fix

- `list_conflicts` (both `DirectClient` and `SdkClient`) tags each returned conflict with its stable `index` into the full report.
- The Web UI resolves by that index rather than the card's position, and keys the card/manual/textarea DOM ids by it too, so the element lookup in `resolveConflict` still resolves after earlier conflicts are resolved.
- `resolve_conflict` rejects an already-resolved index, so a stale index fails loudly instead of deleting something.

## Reproducing

`tests/test_conflict_index_desync.py` fails against unmodified `main` (`f931876`):

```
git checkout origin/main -- memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py
pytest tests/test_conflict_index_desync.py
# FFF  ->  3 failed
```

Two fail because `list_conflicts` exposes no stable identifier for a caller to resolve by (`KeyError: 'index'`); the third because an already-resolved index is silently accepted (`DID NOT RAISE ValueError`).

Before the fix, driving it through the client the way the REST/UI callers do deletes `B_old` when the caller selected conflict C.

## Tests

```
pytest                  # 579 passed, 2 skipped
ruff check .            # All checks passed
ruff format --check .   # 268 files already formatted
mypy memanto            # no issues found in 79 source files
```

Run with a real `MOORCHEH_API_KEY` in `~/.memanto/.env` so the live E2E suite runs rather than skipping.


tharushadp:
Fixed 1 and 3 in `0b23e7d`. Skipping 2 for now, reasoning below.

**1. Version selection before de-duplication.** This was a real gap in the original fix. `_fetch_all_memories` resolved each id to the globally newest version before `search_as_of` applied `created_before`, so an id recreated after `as_of` won de-duplication and was then dropped for being created too late — and the older version that was valid at `as_of` had already been evicted. Net result is the memory disappearing from point-in-time recall entirely, which is worse than the TTL bug this PR started with.

`_fetch_all_memories` now takes an `as_of` parameter and discards versions created after that instant before de-duplicating. `search_as_of` passes `as_of_dt`; default is `None` so `search_recent` and `search_changed_since` are untouched.

Covered by `test_as_of_resolves_id_to_the_version_that_existed_then`: id `addr` is "Lives in Perth" (2028-01-05) and recreated as "Lives in Sydney" (2028-09-01); `as_of=2028-02-01` must return Perth. Against `main` it returns `[]`.

**3. Calendar-dependent test.** Also correct, and the more useful of the two catches. With hard-coded 2026 dates the test only demonstrated anything because the suite happens to run after the fixture expiry — run it earlier in the year and the pre-fix code passes too, so it wasn't a regression guard at all.

Clock is now frozen by an autouse fixture (`FROZEN_NOW = 2029-06-01`, fixtures in 2028), so the outcome no longer depends on the run date. Checked by running the tests against upstream `main`: both bug tests fail, and the control (`test_as_of_still_excludes_memory_that_had_already_expired`) passes on both sides.

**2. Cross-process locking for conflict resolution.** The TOCTOU analysis is right, but I don't think this PR is the place for it.

It's pre-existing rather than introduced here — before this change there was no resolved-state check at all, so a stale index deleted a memory every time. The added guard narrows that window rather than widening it.

Closing it properly means a per-report cross-process lock plus atomic replace, using platform-specific primitives (`msvcrt.locking` on Windows, `fcntl.flock` on POSIX). That changes how conflict reports are persisted and affects `generate_conflict_report` and `list_conflicts` as well, not just the lines here. Rushing a locking protocol into a data-loss fix risks a worse failure than the race it closes — a stale lock file wedging conflict resolution entirely.

Happy to do it as a separate PR if you want it.

Suite after these changes: 458 passed, 24 skipped (E2E, no API key configured at the time) · `ruff check` · `ruff format --check` · `mypy memanto` clean.

tharushadp:
### Disclosure: AI assistance

The BountyHub listing notes that AI tools may be used for assistance but submissions must reflect the submitter's own understanding. Being upfront about how this was produced:

AI tooling was used substantially — reading the codebase, drafting the fixes, and writing the regression tests. I reviewed and approved each change, and I've worked through the reasoning behind both bugs rather than just shipping output. Specifically: `enforce_ttl` defaults to `True` so that no existing caller changes behaviour and only `search_as_of` opts out; and the control test exists because without it the TTL bug could be "fixed" by simply removing expiry enforcement altogether, which would be a regression rather than a fix. Happy to answer questions on any of it.

I'd rather state this plainly than have it come up later.

What doesn't depend on how it was written: the defects are independently checkable. Both regression tests fail against unmodified upstream `main` (`2b2ae9a`):

```
git checkout origin/main -- memanto/app/services/memory_read_service.py
pytest tests/test_as_of_ttl_regression.py
# F.F  ->  2 failed, 1 passed
```

The two failures are the bugs; the pass is the control. `memanto/app/services/memory_read_service.py:531` on `main` is still an unconditional `return self._filter_expired_memories(memories)`, so the point-in-time defect is present today independently of this PR.

tharushadp:
Fixed in `dae9d92`. This one was introduced by this PR, not pre-existing.

When the resolve handlers were switched to the stable full-report index (`cid`), `buildConflictCard` was left creating its DOM ids from the filtered-list position (`i`). Those agree until the first resolution. After that `resolveConflict` calls `getElementById('conflict-' + cid)`, gets `null`, and throws while updating the card — so the user gets a failure toast even though the resolution succeeded, and `invalidateConflictState()` never runs, leaving the badge and history stale.

All three ids (`conflict-`, `manual-`, `manualContent-`) are now keyed by `cid`, and `showManualResolve` receives `cid`. `i` only remains as the `forEach` position feeding the `c.index ?? i` fallback for older backends.

Worth noting the Python suite couldn't have caught this — the UI has no test coverage, so all 496 tests passed either way.

Suite after the change: 496 passed, 0 skipped (including the 24 live E2E tests) · `ruff check` · `ruff format --check` · `mypy memanto` clean.

tharushadp:
@coderabbitai review

tharushadp:
Rebased onto `f931876` and narrowed the scope, so flagging what changed rather than leaving a silent force-push.

The two point-in-time recall bugs this PR originally fixed have since been fixed upstream — `d6bb706` (`search_as_of` recalls since-expired memories) and `79f9675` (as-of dedup version-aware), merged via #1733. Same defects, arrived at independently: `filter_expired` where I had `enforce_ttl`, `created_before` where I had `as_of`. Those commits supersede mine, so I've dropped my versions and the accompanying test file rather than carry duplicate logic.

What's left is the conflict-index bug, which is untouched on `main`. `list_conflicts` still returns only unresolved conflicts with no stable identifier, and `resolve_conflict` still indexes the full report, so the REST and UI callers still delete a memory the caller never selected once any earlier conflict has been resolved.

Verified against unmodified `main`:

```
git checkout origin/main -- memanto/cli/client/direct_client.py memanto/cli/client/sdk_client.py
pytest tests/test_conflict_index_desync.py
# FFF  ->  3 failed
```

PR is now 2 commits, +240 −18 across 4 files. Suite: 579 passed, 2 skipped · `ruff check` · `ruff format --check` · `mypy memanto` clean.

PR #1476, opened 2026-07-15
fix(conflicts): preserve resolutions and fail closed on delete errors (#770)

br413:
## Summary
Fixes three chained memory-integrity bugs for bounty #770:

1. **Conflict regeneration data loss** — re-running detect-conflicts or the nightly schedule overwrote ~/.memanto/conflicts/*.json and erased user resolutions (resolved: true).
2. **False-positive conflict resolution** — resolve_conflict() marked conflicts resolved even when delete_memory() returned False (memories still present).
3. **Phantom session summaries** — POST /{agent_id}/remember hardcoded status: queued and logged failed backend uploads into session Markdown, corrupting daily summaries and conflict detection.

### Bug 1 — resolution wipe
1. Create a conflict report with one entry and set resolved: true.
2. Run generate_conflict_report() again for the same agent/date.
3. Before: resolved entry gone. After: resolution preserved.

### Bug 2 — false resolution
1. Mock delete_memory() to return False.
2. Call resolve_conflict(..., action=keep_new).
3. Before: status resolved in JSON. After: status failed, resolved false.

### Bug 3 — phantom summary
1. Mock store_memory() to return {status: failed, id: m1}.
2. POST to /remember.
3. Before: HTTP 200 + status queued + session MD updated. After: HTTP 502, no phantom log.

## Validation
- python -m pytest tests/test_conflict_integrity.py -q → 4 passed
- python -m ruff check on touched files → clean

/claim #770

## Summary by CodeRabbit

- **Bug Fixes**
  - Memory entries are now logged locally only after successful backend uploads.
  - Upload failures return clearer error responses, and reported statuses reflect backend results.
  - Regenerating conflict reports now preserves previously resolved conflicts.
  - Conflict resolution stops safely when deletion fails and reports the failure instead of marking conflicts resolved.
  - Manual replacements are created only after earlier resolution steps succeed.

- **Tests**
  - Added coverage for conflict report preservation and successful or failed conflict resolution.

PR #1731, opened 2026-07-30
fix(conflicts): resolving a conflict from the Web UI deletes another conflict's memories (#770)

Vyacheslav-Tomashevskiy:
## Summary

Resolving a conflict from the Web UI deletes the memories of a **different conflict** as soon as one earlier conflict in that day's report has been resolved — including a memory that an earlier `keep_new` decision deliberately kept. Deletion is a hard delete, so the loss is not recoverable, and the conflict the user actually judged stays unresolved and comes back on the next reload (click it again → next wrong deletion).

Root cause is a contract asymmetry, not a typo:

* `list_conflicts()` returns **only the unresolved** conflicts, re-indexed `0..n-1`.
* `resolve_conflict(conflict_index=...)` indexes the **full** report on disk (`~/.memanto/conflicts/{agent}_{date}_conflicts.json`).

The index that identifies a conflict never travels with the data, so every consumer that numbers the list response is wrong by the number of already-resolved conflicts ahead of it. Affected consumers today:

| Consumer | Behaviour |
|---|---|
| Web UI (`memanto/app/ui/static/index.html`) | `conflicts.forEach((c, i) => …)` → posts `i` as `conflict_index` — **destructive** |
| `POST /api/v2/agents/{id}/conflicts/resolve` | field documented only as "Conflict index to resolve", so an API/SDK caller pairing it with `GET …/conflicts` is wrong the same way |
| `examples/crewai-memory/run_contradiction.py`, `examples/langgraph-memanto/…/run_contradiction.py` | both list conflicts, print them `1..n`, then resolve `conflict_index=0` — the documented pattern is the buggy one |
| `memanto conflicts` (interactive CLI) | **correct** — it re-reads the report itself and maps display order → report index |

## Reproduction

Driven through the real UI endpoints with a fake `$HOME`, `TestClient(client=("127.0.0.1", …))` to satisfy `_require_local`, and only `delete_memory`/`store_memory` mocked (nothing above that boundary is stubbed). Report: index 0 already resolved with `keep_new` (so `mem-new-0` is the memory the user chose to keep), indices 1 and 2 open.

```
UI list  -> 2 unresolved:
   card position 0: 'Favourite editor'  (old=mem-old-1, new=mem-new-1)
   card position 1: 'Deploy target'     (old=mem-old-2, new=mem-new-2)

User clicks 'Remove Both' on card 0 = 'Favourite editor' -> POST conflict_index=0

main     resolve HTTP 200 {'deleted_old': 'mem-old-0', 'deleted_new': 'mem-new-0', 'status': 'resolved'}
         memories actually deleted:        ['mem-old-0', 'mem-new-0']   <- never shown to the user
         memories the user asked to delete: ['mem-old-1', 'mem-new-1']
         survivor of the earlier keep_new decision destroyed:  True
         conflict the user judged still unresolved:            True

this PR  resolve HTTP 200 {'deleted_old': 'mem-old-1', 'deleted_new': 'mem-new-1', 'status': 'resolved'}
         memories actually deleted: ['mem-old-1', 'mem-new-1']          <- exactly the reviewed pair
```

The API answers `200` with a `deleted_old`/`deleted_new` pair the caller never saw, and the UI shows “Conflict #0 resolved” — nothing surfaces the mistake.

I also executed the **shipped inline JS** (extracted from `index.html`, DOM stubbed, `api()` recorded) instead of reading it:

```
main     card 0 -> DOM id conflict-0 -> POST {conflict_index: 0}                       (no guards)
this PR  card 0 -> DOM id conflict-1 -> POST {conflict_index: 1,
                                              expected_old_memory_id: 'mem-old-1',
                                              expected_new_memory_id: 'mem-new-1'}
```

A second, independent trigger for the same crash exists even with correct indices: the UI’s **“Detect Conflicts Now”** rewrites the report, so a tab that listed conflicts before the rescan resolves a stale index into an entirely different list. That is why the fix is not only “pass the right number”.

## Fix

* `list_conflicts()` (both `DirectClient` and `SdkClient`) tags every returned conflict with `conflict_index`, its position in the full report. Always recomputed from position — the report is generated from LLM output, so a stray `conflict_index` key in the file must never be able to redirect a destructive resolve. Returned dicts are shallow copies, so the field cannot leak back into the report file.
* `resolve_conflict()` **fails closed before deleting anything**:
  * the addressed conflict is already resolved → `ValueError` (previously it happily re-deleted and re-marked it);
  * optional `expected_old_memory_id` / `expected_new_memory_id` don’t match the stored conflict → `ValueError`. This catches index drift from *any* cause, including the regenerated-report race.
* Both guards live in one new module, `memanto/app/utils/conflicts.py`, so the two client copies of this flow cannot drift apart.
* Callers updated to address the report by `conflict_index` and pass the guards: Web UI (`index.html`), `POST /api/ui/conflicts/resolve`, `ConflictResolveRequest` + route docstring, TypeScript SDK (`expectedOldMemoryId` / `expectedNewMemoryId`), both examples. The interactive CLI keeps its own (already correct) mapping and additionally sends the guards, which now also protects it against the report changing between listing and resolving.
* `sdks/typescript/openapi.json` regenerated (`scripts/generate_openapi.py`) so the `openapi-drift` job stays green.

Both new parameters are optional, so existing API/SDK callers keep working; a caller that sends only an index is protected from the “already resolved” case and gets full protection once it forwards the ids.

## Verification

```
pytest                       main: 452 passed, 24 skipped     this PR: 493 passed, 24 skipped
```

41 new cases in `tests/test_conflict_resolution_index.py`, parametrised over `DirectClient` and `SdkClient` and over every destructive action (`keep_old`, `keep_new`, `remove_both`, `manual`): index/report agreement, stray-`conflict_index` override, listing does not mutate the report, indices stable across resolutions, already-resolved refusal, id-mismatch refusal (each side alone too), regenerated-report race, manual refusal storing no orphan memory, the three UI endpoints end-to-end, the request model, the interactive CLI call arguments, and a static guard on `index.html` (the one file CI has no JS harness for).

Mutation testing — reverting only the logic while keeping the tests:

```
revert both halves (pre-fix behaviour)          35 failed / 5 passed
   ▸ the whole existing suite stays green: 452 passed, 24 skipped
   ▸ nothing in the repo covered this
revert guards only (index correct)              23 failed / 17 passed
revert index only (guards on)                   14 failed / 26 passed
```

CI gates run locally with the project’s pinned tool versions: `ruff check .` clean, `ruff format --check .` clean (ruff 0.14.14 per `pyproject.toml`; a newer 0.16 also wants to reformat 11 pre-existing `docs/*.md` code blocks — unrelated), `mypy .` → *Success: no issues found in 107 source files*, `node --check` on the extracted inline JS, and the TypeScript gate end to end: `openapi:generate` → `typecheck` → `build` → `test` (38 passed) with no OpenAPI drift.

Environment: Linux x86-64, Python 3.12.3, Node 22.22, pytest 9.x.

## Deliberately not changed / not covered

* **No Moorcheh API key here**, so nothing ran against the live cloud backend. Everything I touched sits above the storage boundary (report file parsing, index contract, guard, route/UI plumbing), and the tests mock exactly `delete_memory` / `store_memory`. The delete call itself is unchanged.
* **Deletion is still irreversible.** I added no undo or soft-delete: the supersession/trust fields were removed from the active schema on 2026-06-29 (`memanto/app/legacy/REMOVED.md`) and reintroducing them is a product decision, not a bug fix.
* **Behaviour change:** a settled conflict can no longer be re-resolved. I found no caller that does this on purpose — the CLI, the UI and the API list only unresolved conflicts — but it is a deliberate fail-closed choice.
* **Conflict *detection* is untouched.** The report is LLM-generated; I did not change what gets flagged or how `old_memory_id`/`new_memory_id` are chosen.
* **Pre-existing, left alone:** `resolve_conflict()` records a failed `delete_memory` as `result_details["warning"]` and still marks the conflict resolved (PR #1476 covers that path); `list_conflicts()` assumes the report file is a JSON list and will raise on any other shape; the interactive CLI still parses the report itself instead of consuming the new contract field — correct today, but a third copy of the same knowledge.
* The `#N` badge on each conflict card now shows the report index instead of the card position — a visible (and now truthful) change.

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved conflict resolution accuracy by using the authoritative conflict index.
  * Added fail-closed safeguards that reject stale or mismatched resolutions before any deletion.
  * Prevented incorrect conflict resolutions when conflict lists/reports change, including safer concurrency handling to avoid double-resolves.

* **Enhancements**
  * Added optional expected old/new memory identifiers to API, UI, CLI, and TypeScript requests.
  * Updated conflict displays and interactive tools to resolve the exact conflict shown.
  * Expanded API/UI documentation and added concurrency/locking regression tests.

### DIFF
--- a/examples/crewai-memory/run_contradiction.py
+++ b/examples/crewai-memory/run_contradiction.py
@@ -116,11 +116,16 @@ def main() -> None:
             print(
                 "Step 6: Resolving conflict (keeping the newer, higher-confidence value)..."
             )
+            # list_conflicts() returns unresolved conflicts only, so address the
+            # report by the conflict_index it hands back — never by position.
+            target = conflicts[0]
             resolve_result = client.resolve_conflict(
                 agent_id=AGENT_ID,
                 date=today,
-                conflict_index=0,
+                conflict_index=target.get("conflict_index", 0),
                 action="keep_new",
+                expected_old_memory_id=target.get("old_memory_id"),
+                expected_new_memory_id=target.get("new_memory_id"),
             )
             print(f"  Resolution: {resolve_result.get('status', 'resolved')}")
         else:
--- a/examples/langgraph-memanto/memanto_base_store/run_contradiction.py
+++ b/examples/langgraph-memanto/memanto_base_store/run_contradiction.py
@@ -111,12 +111,17 @@ async def main() -> None:
         print()
 
         # Step 6: Resolve by keeping the newer memory.
+        # list_conflicts() returns unresolved conflicts only, so address the
+        # report by the conflict_index it hands back — never by position here.
         print("Step 6: Resolve conflict (keep_new - the recent stomach issue)...")
+        target = conflicts[0]
         resolved = client.resolve_conflict(
             agent_id=agent_id,
             date=today,
-            conflict_index=0,
+            conflict_index=target.get("conflict_index", 0),
             action="keep_new",
+            expected_old_memory_id=target.get("old_memory_id"),
+            expected_new_memory_id=target.get("new_memory_id"),
         )
         print(f"  Resolution status: {resolved.get('status', 'resolved')}\n")
 
--- a/memanto/app/models/__init__.py
+++ b/memanto/app/models/__init__.py
@@ -187,7 +187,16 @@ class ExtractMemoriesRequest(BaseModel):
 class ConflictResolveRequest(BaseModel):
     """Request body for resolving a conflict"""
 
-    conflict_index: int = Field(..., ge=0, description="Conflict index to resolve")
+    conflict_index: int = Field(
+        ...,
+        ge=0,
+        description=(
+            "Index of the conflict to resolve. Use the 'conflict_index' field "
+            "of the entry returned by GET /conflicts — that response contains "
+            "only unresolved conflicts, so its own ordering is not the index "
+            "the report is addressed by."
+        ),
+    )
     action: Literal["keep_old", "keep_new", "keep_both", "remove_both", "manual"] = (
         Field(
             ...,
@@ -204,6 +213,18 @@ class ConflictResolveRequest(BaseModel):
     )
     manual_type: str | None = Field(
         None, description="Optional memory type for manual action"
+    )
+    expected_old_memory_id: str | None = Field(
+        None,
+        description=(
+            "Optional guard: the old_memory_id the caller reviewed. When set, "
+            "the resolution is rejected (400) before any memory is deleted if "
+            "the stored conflict does not match."
+        ),
+    )
+    expected_new_memory_id: str | None = Field(
+        None,
+        description="Optional guard for new_memory_id, see expected_old_memory_id.",
     )
 
     @model_validator(mode="after")
--- a/memanto/app/routes/memory.py
+++ b/memanto/app/routes/memory.py
@@ -1149,6 +1149,12 @@ async def resolve_conflict(
     Resolve a conflict for an agent.
 
     Uses the same underlying conflict resolution service used by CLI.
+
+    ``conflict_index`` is the ``conflict_index`` field of an entry returned by
+    ``GET /{agent_id}/conflicts``, not that response's own ordering (it lists
+    unresolved conflicts only). Pass ``expected_old_memory_id`` /
+    ``expected_new_memory_id`` to have the call rejected before any deletion if
+    the index no longer addresses the reviewed conflict.
     """
     enforce_session_scope(session, agent_id)
 
@@ -1163,6 +1169,8 @@ async def resolve_conflict(
             request.action,
             request.manual_content,
             request.manual_type,
+            request.expected_old_memory_id,
+            request.expected_new_memory_id,
         )
         return {
             "agent_id": agent_id,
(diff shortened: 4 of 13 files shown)


Vyacheslav-Tomashevskiy:
Fixed in 364e94a — the finding was right, and it was the one gap left: the guards fail closed, but they were not *atomic*.

`conflict_report_lock(json_path)` now wraps the whole transaction (reload → validate → delete → persist) in both clients, using an advisory `flock` on a sidecar `.lock` file so it holds across processes — a CLI resolve racing the Web UI, not just two threads. Platforms without `fcntl` fall back to no locking rather than failing the resolve; the in-transaction guards there are unchanged, they are just not serialized.

Proof the test is sensitive to it: with `conflict_report_lock` stubbed to a no-op, two racing resolves of conflict 1 give `recorder.deleted == ['mem-old-1', 'mem-old-1']` and the second write clears the first `resolved` marker; with the lock, the loser is refused with *"already resolved"* and the memory is deleted once. Also covered: the lock is exclusive while held and released afterwards, and the sequential two-resolve path is unchanged.

`498 passed, 24 skipped`; `ruff`, `ruff format --check` and `mypy` clean.

Xenogents:
After reviewing your PR, we've determined that the issue has already been addressed by another contribution (or was already fixed in the codebase). Because of this, we're unable to merge this PR. #1728

PR #1743, opened 2026-07-31
fix: prevent FastAPI Header metadata from becoming an API key

Roezna:
## Summary

Fix the backend dependency wrappers so ordinary Python calls use the configured Moorcheh API key instead of forwarding FastAPI's `Header` metadata object as a credential.

## Flaw and impact

`get_moorcheh_client()` and `get_async_moorcheh_client()` used `Header(None, alias="X-Api-Key")` as their real Python default. That works when FastAPI resolves the dependency, but internal code also invokes the functions directly. In a plain call, the default is a truthy `fastapi.params.Header` instance rather than `None`.

The cloud dispatcher therefore selects that object as `key_to_use` and constructs a Moorcheh client with invalid credentials. Direct call sites include the session-scoped answer and upload-file routes, so correctly configured deployments can fail authentication for core operations even though `settings.MOORCHEH_API_KEY` is valid.

Minimal pre-fix reproduction:

```python
from memanto.app.clients.moorcheh import get_moorcheh_client

default = get_moorcheh_client.__defaults__[0]
assert type(default).__name__ == "Header"  # invalid credential forwarded
```

## Fix

Use FastAPI's `Annotated[..., Header(...)]` form while retaining `None` as the actual Python default. FastAPI still reads `X-Api-Key`, explicit embedded callers still pass their key, and direct calls now correctly fall back to the configured key.

Regression coverage verifies:

- sync direct calls dispatch `api_key=None`
- async direct calls dispatch `api_key=None`
- explicit sync and async keys remain unchanged

## Validation

- `python -m pytest tests/test_backend.py -q` - 19 passed
- `python -m pytest -q` - full non-live suite passed; 24 live E2E tests skipped because they require a real Moorcheh key
- `python -m ruff check memanto/app/clients/moorcheh.py tests/test_backend.py` - passed
- `python -m mypy memanto --no-error-summary` - passed

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API-key handling for direct synchronous and asynchronous client access.
  * Prevented internal request-header metadata from being treated as an API key during direct calls.
  * Ensured explicitly provided API keys and request-header API keys are forwarded correctly.

* **Tests**
  * Added coverage for missing, explicitly supplied, and request-header API-key scenarios across both client access modes.

### DIFF
--- a/memanto/app/clients/moorcheh.py
+++ b/memanto/app/clients/moorcheh.py
@@ -10,7 +10,7 @@ Service code keeps calling ``get_moorcheh_client()`` and uses the same
 backends expose it.
 """
 
-from typing import Any
+from typing import Annotated, Any
 
 from fastapi import Header
 from moorcheh_sdk import AsyncMoorchehClient, MoorchehClient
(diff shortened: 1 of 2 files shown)


Roezna:
Addressed the review nit in cfbb4a4: added a route-level FastAPI regression test proving that `X-Api-Key` is injected correctly through both dependency wrappers. Focused backend suite is now 20/20 passing; full non-live suite, Ruff, and mypy remain clean.

Xenogents:

PR #1759, opened 2026-08-01
fix: guard idempotency store with a lock

Rerzx:
### Root cause
`IdempotencyStore` uses an in-memory dict without synchronization. The check-then-write flow allows concurrent requests with the same key to bypass the check and duplicate writes.

### Fix
- Added a `threading.Lock` to guard `get_record`, `store_record`, and `get_stats`.
- `_cleanup_expired` now assumes the lock is held, preventing races during cleanup.

### Files
- `memanto/app/legacy/idempotency.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when handling duplicate requests concurrently.
  * Added clear conflict responses when an identical request is already being processed.
  * Ensured completed requests return their previously stored responses.
  * Prevented race conditions during record cleanup, retrieval, storage, and statistics operations.
  * Preserved existing expiration and statistics behavior.

### DIFF
--- a/memanto/app/legacy/idempotency.py
+++ b/memanto/app/legacy/idempotency.py
@@ -5,9 +5,9 @@ Idempotency Handling for MEMANTO
 import hashlib
 import re
 import time
+import threading
 from dataclasses import dataclass
 from typing import Any
-
 from fastapi import HTTPException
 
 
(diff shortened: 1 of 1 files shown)


Xenogents:
We are closing this PR because the idempotency features are all legacy and so we will avoid making changes to it.

PR #1760, opened 2026-08-01
fix(cli): restore minimum-confidence recall filtering

atlascraft-codex:
## Problem

The default CLI client (`SdkClient`) dropped the `min_confidence` argument even though `MemoryReadService.search_memories` and the parallel `DirectClient` both support it. The CLI also had no `--min-confidence` option.

Reproduction before this change:

```text
memanto recall "reliable facts" --min-confidence 0.8
Error: No such option: --min-confidence
```

Calling `SdkClient.recall(..., min_confidence=0.8)` raised `TypeError: unexpected keyword argument 'min_confidence'`. Users therefore could not exclude low-confidence memories through the primary CLI/SDK path, despite the filtering implementation already existing below it.

## Fix

- Add a bounded `--min-confidence` CLI option (`0.0` through `1.0`).
- Forward the value through the CLI command.
- Restore `SdkClient` parity with `DirectClient` and pass the value to `MemoryReadService`.
- Add regressions at both boundaries: CLI → client and SDK client → read service.

The retrieval/filtering algorithm is unchanged; this reconnects the public path to the existing implementation.

## Verification

- New reproduction tests fail on `main` with the two errors above and pass with this patch.
- `python -m pytest -q` — full suite passed; only the existing live-key/POSIX-only skips remain.
- `python -m ruff check ...` — passed.
- `python -m ruff format --check ...` — passed.
- `python -m mypy memanto/cli/client/sdk_client.py memanto/cli/commands/memory.py` — passed.

Relates to #770.

## Summary by CodeRabbit

* **New Features**
  * Added an optional minimum confidence filter to memory searches, helping narrow results to memories meeting a chosen confidence threshold.
  * Added the `--min-confidence` option to the recall command, accepting values from 0.0 to 1.0.
  * Updated the SDK recall API to support confidence-based filtering.

* **Tests**
  * Added coverage verifying confidence filters are accepted, forwarded, and correctly handled when returning memories.


atlascraft-codex:
Follow-up verification after review:

- Addressed the positional-argument compatibility finding in `30dcf65`: `min_confidence` now follows the existing datetime parameters, with a regression test covering legacy positional calls.
- Re-ran the full test suite, Ruff, formatting, and mypy successfully.
- Ran a read-only live Moorcheh check through the updated `SdkClient` on an existing test agent. The same query returned 10 memories at `min_confidence=0.0` (stored confidences: 0.90 and 0.95) and 0 memories at `min_confidence=1.0`, confirming the filter reaches and is enforced by the live service path. No data was created or modified.

PR #1763, opened 2026-08-01
fix(services): bound conflict detection query length to prevent embedding context overflow [Fixes #1329]

### LINKED ISSUE #1329: `detect-conflicts` (on-prem) fails on any active day because the conflict query exceeds the embedding model's context window

gianmarcozap:
## Summary\nFixes **Issue #1329** where \detect-conflicts\ fails with HTTP 400 on active days because \generate_conflict_report\ passed un-truncated \conflict_prompt\ directly as the retrieval \query\ token embedding.\n\n## Changes Made\n- Applied \_truncate_embedding_query(full_text, model=get_active_embedding_model())\ to bound the retrieval embedding query.\n- Moved the prompt instructions to \header_prompt\ in \client.answer.generate\.\n\n## Payout Target Wallet\n/claim 0xBd6B1B6118eC9D736EE1d5E476f86BCA1b3739f5

### DIFF
--- a/memanto/app/legacy/idempotency.py
+++ b/memanto/app/legacy/idempotency.py
@@ -147,7 +147,25 @@ class IdempotencyHandler:
             # Return cached response
             return record.response
 
-        return None
+    @staticmethod
+    def reserve_or_get_idempotent_response(
+        idempotency_key: str | None,
+        memory_id: str = "",
+        response: dict[str, Any] | None = None,
+        ttl_seconds: int = 86400,
+    ) -> tuple[bool, dict[str, Any] | None]:
+        """Atomic reservation handler for production write idempotency"""
+        if not idempotency_key:
+            return True, None
+
+        response_dict = response if response is not None else {}
+        is_owner, record = idempotency_store.reserve_or_get(
+            idempotency_key=idempotency_key,
+            memory_id=memory_id,
+            response=response_dict,
+            ttl_seconds=ttl_seconds,
+        )
+        return is_owner, record.response
 
     @staticmethod
     def store_idempotent_response(
@@ -156,7 +174,7 @@ class IdempotencyHandler:
         response: dict[str, Any],
         ttl_seconds: int = 86400,
     ):
-        """Store response for idempotency"""
+        """Store response for idempotency thread-safely"""
         if not idempotency_key:
             return
 
@@ -193,7 +211,7 @@ class IdempotencyHandler:
 
 
 def handle_write_idempotency(idempotency_key: str | None) -> dict[str, Any] | None:
-    """Handle idempotency for write operations"""
+    """Handle idempotency for write operations thread-safely"""
     if not idempotency_key:
         return None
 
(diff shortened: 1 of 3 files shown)


gianmarcozap:
Closing in favor of a clean PR scoped strictly to Issue #1329 (no unrelated idempotency changes). See replacement PR.

PR #1807, opened 2026-08-03
fix: prevent silent data loss in search_changed_since when memory id is None

windy202508:
### Summary

In `MemoryReadService.search_changed_since`, memories with a missing or `None` id are **silently dropped** after the first one, because the dedup logic treats all `None`-id memories as duplicates of each other.

### Root Cause

```python
seen_ids = set()
for memory in all_memories:
    mem_id = memory.get("id")
    if mem_id in seen_ids:    # ← None in seen_ids is True after first None-id memory
        continue               # ← ALL subsequent None-id memories are skipped!
    seen_ids.add(mem_id)       # ← adds None to the set
```

After the first memory with `id=None`, `seen_ids` contains `None`. Every subsequent memory with `id=None` hits `if mem_id in seen_ids` → `True` and is silently skipped.

### Impact

- **Data loss**: If a Moorcheh namespace contains multiple documents without an `id` field, `search_changed_since` returns only the first one, silently dropping the rest.
- This can happen with documents written outside Memanto's own store path (manual test data, other tools sharing the namespace), or with documents whose `id` field is missing/corrupt.
- The bug is in a **differential retrieval** endpoint (`recall/changed-since`), meaning users checking "what changed recently" get incomplete results with no error or warning.

### Fix

Only add to `seen_ids` and check for duplicates when `mem_id is not None`. Memories without an id are never considered duplicates of each other:

```python
seen_ids: set[Any] = set()
for memory in all_memories:
    mem_id = memory.get("id")
    if mem_id is not None and mem_id in seen_ids:
        continue
    if mem_id is not None:
        seen_ids.add(mem_id)
```

### Test Plan

Added `tests/test_changed_since_none_id_dedup.py` with three regression tests:
1. `test_multiple_none_id_memories_are_all_kept` — all None-id memories appear in results
2. `test_real_id_dedup_still_works` — real-id dedup is preserved
3. `test_mixed_none_and_real_ids` — mix of None and real ids behaves correctly

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed changed-memory searches incorrectly removing separate memories that do not have an ID.
  * Duplicate memories with the same valid ID continue to be consolidated.
* **Tests**
  * Added coverage for ID-less memories, duplicate IDs, and mixed results.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -439,13 +439,18 @@ class MemoryReadService:
 
             # Filter to only changed memories
             changed_memories = []
-            seen_ids = set()
+            seen_ids: set[Any] = set()
 
             for memory in all_memories:
                 mem_id = memory.get("id")
-                if mem_id in seen_ids:
+                # Only dedup by id when the id is present. Memories with a
+                # missing or None id are not duplicates of each other --
+                # treating them as such silently drops all but the first
+                # id-less memory (data loss, bounty #770).
+                if mem_id is not None and mem_id in seen_ids:
                     continue
-                seen_ids.add(mem_id)
+                if mem_id is not None:
+                    seen_ids.add(mem_id)
 
                 # Check if created after since_date (new memory)
                 created_at = memory.get("created_at")
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -618,6 +618,9 @@ class MemoryReadService:
                     break
 
         latest_by_id: dict[str, tuple[tuple[datetime, int], dict[str, Any]]] = {}
+        # Memories without an id are kept as-is (no dedup possible).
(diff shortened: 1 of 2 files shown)


windy202508:
Updated to also fix  per CodeRabbit's review.

The original fix only addressed the dedup logic in , but  was dropping all memories with /missing uid=197108(Administrator) gid=197121 groups=197121 at line 622 () before they could even reach .

**Changes in this commit:**
- Added  list to  to collect memories with no id instead of dropping them
- These id-less memories are included in the final results without dedup (since dedup is impossible without an id)
- The existing dedup logic for memories with valid ids remains unchanged

This ensures the full pipeline preserves id-less memories end-to-end:  →  → results.

windy202508:
Closing in favor of #1817 which includes regression tests for this bug.

PR #1808, opened 2026-08-04
fix: Moorcheh filter injection, pagination truncation, session-renew lockout, stale-field persistence (Bounty #770)

liushazyy:
## Bounty #770 — 4 deterministic bugs fixed

This PR fixes four reproducible bugs in the Memanto core package (all verified with failing-then-passing tests). Each fix is minimal, single-purpose, and covered by a regression test in `tests/failing_tests/`.

### 1. HIGH — Moorcheh query-language injection via free-text `#` tokens
`MemoryReadService._build_filtered_query` concatenates the caller's free-text `query` with server-side Moorcheh filters (`#memory_type:`, `#tag`, `#status:`). Moorcheh parses any `#key:value` / `#tag` token embedded in the text, so a query like `"meeting notes #memory_type:preference"` silently injects a `memory_type=preference` filter — narrowing/emptying recall results and **bypassing the API's declared `type`/`tags`/`status` filter contract** (a caller asking for `type=["fact"]` can get an empty set).

**Fix:** escape user-supplied `#` (→ full-width `＃`) before appending server filters; server-side filters remain untouched. `_escape_moorcheh_filter_tokens` + regression test.

### 2. HIGH — silent pagination truncation past the 100-row cap
`search_memories` fetches `top_k = min(max(limit+offset, 100), 100)` ≡ **always 100**. When a namespace holds >100 matching rows, any page with `offset+limit ≥ 100` returns `[]` with `has_more=False`, and `total_found` reports the page length instead of the hit count — callers relying on `offset/limit/has_more` silently lose every result past 100.

**Fix:** fail closed — `offset+limit > MOORCHEH_MAX_TOP_K` now raises a clear `ValueError` (wrapped as `MemoryError` by the service) instead of silently lying.

### 3. HIGH — session auto-renew locks out `X-Session-Token` header clients
`auth_deps.get_current_session` auto-renews sessions ≤30 min before expiry, which **invalidates the caller's current token** (new `session_id`). The renewed token is returned only via `Set-Cookie`; clients that authenticate with the documented `X-Session-Token` header (curl, SDKs, third-party integrations) get **no** new token and 401 on their very next request — silently locked out of an "auto-renewing" session.

**Fix:** when the request authenticated via `X-Session-Token`, the renewed token is also returned in the `X-Session-Token` response header.

### 4. MED — `update_memory` persists stale flat fields (incl. search `score`) into storage
`update_memory` copies every key of the *formatted* existing record (`title/content/type/score/tags/...`) not present in the new document into the stored doc. The previous search's relevance `score` and pre-edit flat values are persisted and later observed by consumers reading flat fields — stale data corruption, deterministic.

**Fix:** copy forward only schema-external keys (e.g. `original_id`); exclude schema-internal/transient keys.

---

### Reproducibility
`python tests/failing_tests/test_bounty_770_injection_pagination.py` — 6/6 pass on this branch; the first two tests fail against `main` (injection not escaped; pagination silently truncates).

### Files changed
- `memanto/app/services/memory_read_service.py` — fix 1 & 2
- `memanto/app/routes/auth_deps.py` — fix 3
- `memanto/app/services/memory_write_service.py` — fix 4
- `tests/failing_tests/test_bounty_770_injection_pagination.py` — new regression tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved search safety, pagination limits, result totals, and tag handling.
  * Header-authenticated sessions now receive renewed session tokens in the response.
  * Memory updates no longer preserve outdated or internal metadata.
  * Improved validation and error messages for sessions, imports, API keys, and configuration.
  * Conflict reports and exports now honor configured data directories and use UTC timestamps.
  * Large conversation content is handled more reliably during memory creation.
* **Tests**
  * Added broad regression coverage for search, sessions, exports, migrations, integrations, and conflict handling.

### DIFF
--- a/integrations/langgraph/langgraph_memanto/nodes.py
+++ b/integrations/langgraph/langgraph_memanto/nodes.py
@@ -246,6 +246,16 @@ def create_remember_node(
             return {"messages": []}
 
         content = "\n\n".join(messages_to_remember)
+        # The memory store caps content at InputLimits.MAX_TEXT_LENGTH (10k
+        # chars) and rejects anything larger with a ValueError. Truncate here
+        # (keeping the newest text, i.e. the end of the joined content) so a
+        # long pasted document or code block is stored instead of silently
+        # dropped — the previous behavior swallowed the ValueError in the
+        # except below, logged an error, and returned {"messages": []} as if
+        # the write had succeeded.
+        max_content = 10_000
+        if len(content) > max_content:
+            content = content[-max_content:]
         title = content if len(content) <= 50 else content[:47] + "..."
 
         agent_client, agent_lock = _cache.get(resolved_agent_id)
--- a/integrations/langgraph/langgraph_memanto/store.py
+++ b/integrations/langgraph/langgraph_memanto/store.py
@@ -403,8 +403,32 @@ class MemantoStore(BaseStore):
             logger.warning("MemantoStore: Failed to list agents: %s", e)
             return []
 
+        # SdkClient.list_agents returns the AgentList envelope dict
+        # ({"agents": [...], "count": N, "warnings": [...]}), not a bare
+        # list. Iterating the dict directly would yield its string keys and
+        # crash on .get(); unwrap the list defensively so both the current
+        # envelope shape and a hypothetical bare-list backend work.
+        if isinstance(agents, dict):
+            agent_items = agents.get("agents")
+            if not isinstance(agent_items, list):
+                logger.warning(
+                    "MemantoStore: unexpected list_agents payload keys: %s",
+                    sorted(agents),
+                )
+                return []
+        elif isinstance(agents, list):
+            agent_items = agents
+        else:
+            logger.warning(
+                "MemantoStore: unexpected list_agents payload type: %s",
+                type(agents).__name__,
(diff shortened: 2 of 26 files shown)


liushazyy:
已按 CodeRabbit 提示补齐 4 个 docstring（commit 54e5327），docstring 覆盖率已达标。6 个回归测试全部通过。

liushazyy:
Understood — consolidated everything into this single PR per your note. The branch now contains all rounds (2-8) on top of round 1: 35 bugs total, each with a failing-then-passing regression test in tests/failing_tests/. Full suite: 624 passed, 26 skipped. Please let me know if you want the commits squashed.

Xenogents:
The bounty winner has already been determined but we will still continue to be merging useful bug fixes into the repository. Your fixes have been cherry picked and implemented into the repository.

liushazyy:
Thank you for the kind words and for cherry-picking the fixes into the codebase — glad they could be useful regardless of the bounty outcome. It was a great exercise digging into Memanto’s internals, and I’ll be happy to contribute more fixes in the future. Thanks again to the team for the thorough review process.

PR #1810, opened 2026-08-04
fix: prevent silent data loss when content paragraphs start with "Tags: "

windy202508:
### Problem
In memory_read_service.py, the _format_memory_item method strips the trailing Tags: block from recalled memory content. The current check is too loose - it only verifies that the last paragraph starts with "Tags: ". This causes silent data loss when a legitimate content paragraph happens to start with "Tags: ".

### Reproduction
1. Store a memory with content: "Tags: are useful for categorization" and tags ["python", "ai"]
2. The serializer produces: [FACT] title

Tags: are useful for categorization

Tags: python, ai
3. On readback, rpartition finds the last paragraph, and startswith("Tags: ") matches the content paragraph
4. The content paragraph is silently stripped, resulting in empty or truncated content

### Impact
- Data loss: Content paragraphs starting with "Tags: " are permanently truncated on readback
- Affects any memory about tagging systems, label taxonomies, or content categorization
- No error or warning - the data simply disappears

### Fix
Before stripping the trailing Tags: block, verify that the tags listed in it actually match the record tags from metadata. This prevents false positives.

### Test
Added tests/test_format_memory_tags_bug.py with three regression tests.

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory item formatting to better preserve legitimate content paragraphs that start with "Tags:" while correctly removing only auto-generated tag blocks.

* **Tests**
  * Added regression tests for memory tag formatting to ensure proper content preservation and tag handling.

### DIFF
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -113,7 +113,10 @@ class MemoryRecord(BaseModel):
         # Format text as standardized card for semantic search
         text = f"[{memory_type.upper()}] {self.title}\n\n{self.content}"
         if self.tags:
-            text += f"\n\nTags: {', '.join(self.tags)}"
+            text += f"
+
+<!--memanto-tags:v1-->
+Tags: {, '.join(self.tags)}"
 
         # Build document with flat metadata fields (not nested!)
         document = {
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -1006,29 +1006,36 @@ class MemoryReadService:
             # stripped instead of leaking into the returned title.
             title_match = re.match(r"^\[.*?\]\s*(.*)$", first_line, re.DOTALL)
             title = title_match.group(1).strip() if title_match else first_line.strip()
+            # Strip ONLY a genuine trailing tags block using a unique
+            # versioned marker. The serializer appends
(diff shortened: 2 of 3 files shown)


windy202508:

### Use a distinguishable serialized tags block ✅

Fixed in both the serializer (`core.py`) and the reader (`memory_read_service.py`):

**Serializer (core.py):** Now appends a unique, versioned marker before the tags block:
```python
# Before: text += f"\n\nTags: {', '.join(self.tags)}"
# After:  text += f"\n\n<!--memanto-tags:v1-->\nTags: {', '.join(self.tags)}"
```

**Reader (memory_read_service.py):** Checks for the marker first:
- If `<!--memanto-tags:v1-->` is present → strip from marker onwards (definitely serialized metadata)
- If marker not found → fall back to old behavior for backward compatibility (check `Tags: ` prefix + verify tags match)

This prevents the reader from stripping legitimate user content that happens to end with `\n\nTags: python, ai` when the record's tags are `["python", "ai"]`.

Refs #770

windy202508:
Closing in favor of #1819 (rebased onto current main to resolve merge conflicts). The fix is the same: using a versioned marker to distinguish serialized tags from user content.

PR #1812, opened 2026-08-04
fix: prevent silent data loss in temporal filter for memories with missing timestamps [Bounty #770]

windy202508:
### Problem

`_apply_temporal_filter` silently dropped memories whose `created_at` field was missing or unparseable when a temporal filter (`created_after` / `created_before`) was active. This caused **timeline amnesia** — memories that should have been kept were silently lost, inconsistent with the documented fail-open behavior in `_filter_expired_memories` and `search_as_of`.

### Reproduction

1. Store a memory without a `created_at` timestamp (can happen when documents are written outside Memanto's own store path — manual test data, other tools sharing the namespace, or corrupt timestamp fields)
2. Call `search_memories(query="...", created_after="2026-01-01T00:00:00Z")`
3. The memory with the missing timestamp is **silently dropped** from the results
4. No error or warning — the data simply disappears

### Root Cause

```python
# Before fix (line ~780):
filtered = []
for result in results:
    raw_created = result.get("created_at")
    if not raw_created:
        continue  # ← BUG: silently drops memories without timestamps

    try:
        created_dt = parse_iso_timestamp(raw_created)
    except (ValueError, AttributeError, TypeError):
        continue  # ← BUG: silently drops memories with corrupt timestamps
```

Both `continue` statements silently drop the memory. This is inconsistent with:
- `_filter_expired_memories`: fails open (keeps memories with unparseable `expires_at`)
- `search_as_of`: fails open (keeps memories with malformed `expires_at`)

### Fix

Fail open: keep memories with missing or unparseable timestamps instead of dropping them. A missing or corrupt timestamp is not proof the memory falls outside the requested time window.

```python
# After fix:
filtered = []
for result in results:
    raw_created = result.get("created_at")
    if not raw_created:
        # Fail open: keep the memory (timeline amnesia prevention)
        filtered.append(result)
        continue

    try:
        created_dt = parse_iso_timestamp(raw_created)
    except (ValueError, AttributeError, TypeError):
        # Fail open: keep the memory
        filtered.append(result)
        continue
    ...
```

### Tests

Added `tests/test_temporal_filter_fail_open.py` with 5 regression tests:
1. `test_missing_created_at_is_kept_with_created_after` — memory with `None` timestamp survives `created_after` filter
2. `test_missing_created_at_is_kept_with_created_before` — memory with `None` timestamp survives `created_before` filter
3. `test_unparseable_created_at_is_kept` — memory with corrupt timestamp string survives filtering
4. `test_valid_timestamps_are_still_filtered_correctly` — normal filtering behavior preserved
5. `test_all_missing_timestamps_are_kept` — all-missing scenario

### Verification

- `python -m pytest tests/test_temporal_filter_fail_open.py -v` — 5 passed
- `python -m pytest tests/test_memory_read_temporal_recall.py tests/test_memory_read_as_of.py tests/test_as_of_expired_recall.py tests/test_as_of_date_only_parsing.py -v` — 25 passed (no regressions)
- `python -m ruff check memanto/app/services/memory_read_service.py tests/test_temporal_filter_fail_open.py` — passed
- `python -m ruff format --check ...` — passed

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Memories with missing or invalid creation dates are now retained during temporal filtering and point-in-time collection.
  * Valid timestamps continue to respect “created after” and “created before” limits.

* **Tests**
  * Added coverage for missing, invalid, and valid creation dates, including collections containing only undated memories.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -776,11 +776,25 @@ class MemoryReadService:
         for result in results:
             raw_created = result.get("created_at")
             if not raw_created:
+                # Fail open: a memory without a created_at timestamp should
+                # not be silently dropped when a temporal filter is active.
+                # The memory may have been written outside Memanto own store
+                # path (manual test data, other tools sharing the namespace)
+                # with a missing or corrupt timestamp field. Dropping it here
+                # causes timeline amnesia - the exact data-loss pattern that
+                # bounty #770 is designed to catch. This is consistent with
+                # _filter_expired_memories, which also keeps memories whose
+                # expires_at cannot be parsed.
+                filtered.append(result)
                 continue
 
             try:
                 created_dt = parse_iso_timestamp(raw_created)
             except (ValueError, AttributeError, TypeError):
+                # Fail open: an unparseable timestamp is not proof the memory
+                # falls outside the requested time window. Keep it rather than
+                # silently dropping it (timeline amnesia).
+                filtered.append(result)
                 continue
 
             if after_dt is not None and created_dt < after_dt:
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -627,13 +627,12 @@ class MemoryReadService:
             # after as_of cannot displace the version valid then (bounty #770).
             if before_dt is not None:
                 raw_created = formatted.get("created_at")
-                if not raw_created:
-                    continue
-                try:
-                    if parse_iso_timestamp(str(raw_created)) > before_dt:
-                        continue
-                except (TypeError, ValueError):
-                    continue
+                if raw_created:
+                    try:
+                        if parse_iso_timestamp(str(raw_created)) > before_dt:
+                            continue
+                    except (TypeError, ValueError):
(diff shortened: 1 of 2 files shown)


windy202508:

### Apply fail-open behavior before point-in-time de-duplication ✅

Fixed in `_fetch_all_memories`: when `created_before` is set, memories with missing or unparseable `created_at` values are now **retained as candidates** instead of being silently dropped.

**Before:**
```python
if before_dt is not None:
    raw_created = formatted.get("created_at")
    if not raw_created:
        continue  # ← silently drops memories with missing timestamps
    try:
        if parse_iso_timestamp(str(raw_created)) > before_dt:
            continue
    except (TypeError, ValueError):
        continue  # ← silently drops memories with bad timestamps
```

**After:**
```python
if before_dt is not None:
    raw_created = formatted.get("created_at")
    if raw_created:
        try:
            if parse_iso_timestamp(str(raw_created)) > before_dt:
                continue
        except (TypeError, ValueError):
            pass  # Fail-open: keep memories with unparseable timestamps
```

This ensures `search_as_of` does not silently exclude memories with unknown timestamps.

Refs #770

Xenogents:

PR #1815, opened 2026-08-05
fix: bound pagination loop, fail-closed temporal limits, truthful total_found, configurable conflict dir (Bounty #770 round 2)

liushazyy:
## Bounty #770 — round 2: 4 more deterministic bugs

Second submission batch on top of PR #1808. All four bugs are reproducible,
covered by failing-then-passing regression tests in
`tests/failing_tests/test_bounty_770_round2.py`.

### 1. HIGH — unbounded pagination loop in `_fetch_all_memories`
`MemoryReadService._fetch_all_memories` iterated `while True` over
`fetch_text_data` pages with no protection against a repeated or
never-ending `next_token`. A storage layer that keeps returning
`has_more=True` with the same (or always-new) token spins unbounded HTTP
fetches. Fix: track seen tokens per namespace and cap at 200 pages
(20k documents), far beyond any real agent memory set.

### 2. MED — temporal recall silently mis-slices with negative/zero limit
`search_as_of` / `search_changed_since` / `search_recent` accepted any
`limit` and applied it with Python slicing. `limit=-1` returned
everything-but-the-last memory; `limit=0` faked an empty result set that
is indistinguishable from "no memories". `search_memories` already fails
closed on invalid limits; these three now do too (ValueError).

### 3. MED — `total_found` lied about match totals
`search_memories` reported `total_found` as the length of the returned
page, so a paged recall with `limit=10` showed `total_found=10` even when
hundreds of memories matched. Paginating clients use `total_found` to
decide "are there more" and render counts. Fix: report the post-filter
match total (consistent with `total_available`).

### 4. LOW — conflict reports ignored `MEMANTO_DATA_DIR`
`DailyAnalysisService.generate_conflict_report` hardcoded
`~/.memanto/conflicts`, so with a configured data dir conflict reports
landed in a different directory than the daily summaries they are meant to
accompany. Fix: `get_data_dir() / "conflicts"`.

### Verification
- 8 new regression tests (all fail on pre-fix code, pass on fixed code)
- Full suite: 592 passed, 2 skipped, 0 failed on latest main

## Summary by CodeRabbit

* **Bug Fixes**
  * Conflict reports and export files now use the configured data directory.
  * Search totals accurately reflect filtered and paginated matches.
  * Invalid time-based search limits and recall settings are rejected safely.
  * Memory retrieval pagination prevents stalled or unbounded searches.
  * Tags containing commas are rejected, preserving consistent tag handling.

* **Tests**
  * Added regression coverage for search, validation, pagination, paths, exports, and tag handling.

### DIFF
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -16,7 +16,18 @@ from memanto.app.constants import (
 )
 
 MemoryTag = Annotated[
-    str, StringConstraints(strip_whitespace=True, min_length=1, max_length=64)
+    str,
+    StringConstraints(
+        strip_whitespace=True,
+        min_length=1,
+        max_length=64,
+        # Commas are the wire-format separator: to_moorcheh_document joins
+        # tags with "," and readers split on ",". A tag containing a comma
+        # is silently corrupted on the write->read round-trip (the tag is
+        # split into two, a phantom tag appears) and makes #tag filtering
+        # match the wrong rows. Reject it at the schema boundary.
+        pattern=r"^[^,]+$",
+    ),
 ]
 BoundedTags = Annotated[list[MemoryTag], Field(max_length=20)]
 
--- a/memanto/app/services/daily_analysis_service.py
+++ b/memanto/app/services/daily_analysis_service.py
@@ -215,7 +215,10 @@ Format the output as a Markdown report:
         validate_safe_id(agent_id, "agent_id")
         validate_safe_id(date, "date")
 
-        conflicts_dir = Path.home() / ".memanto" / "conflicts"
+        # Use the configured data dir (same as summaries) instead of a
+        # hardcoded ~/.memanto path, so MEMANTO_DATA_DIR is honored
+        # consistently across generated artifacts.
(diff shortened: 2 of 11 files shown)


liushazyy:
Round 5 pushed (02d41aa): addressed all coderabbitai review points.

1. ValueError contract preserved - limit validation moved to a _validate_limit helper called OUTSIDE the try/except wrappers of search_as_of / search_changed_since / search_recent, so an invalid limit raises ValueError directly instead of being swallowed into MemoryError, and it fires even when no namespaces exist (no more empty-result early return bypassing validation).
2. Global pagination budget - pages_fetched moved outside the namespace loop in _fetch_all_memories, so the 200-page bound applies across ALL namespaces, not per-namespace.
3. Tests hardened - limit tests now assert ValueError (not bare Exception); added runtime test test_cli_conflicts_reads_configured_data_dir_at_runtime that points get_data_dir at a temp root and proves the CLI loads its report from the configured path.

All 20 bounty-770 tests pass plus 42 memory-read tests, no regressions.

Xenogents:

PR #1817, opened 2026-08-05
fix: silent data loss in search_changed_since when memory id is None

windy202508:
### Problem
In `search_changed_since`, the `seen_ids` set is used to deduplicate memories by their `id`. However, when a memory lacks an `id` field (returns `None` from `.get('id')`), the dedup logic **silently drops all subsequent memories that also lack an id**.

### Root Cause
```python
mem_id = memory.get('id')
if mem_id in seen_ids:   # None is a valid set member!
    continue             # silently skips ALL id=None memories after the first
seen_ids.add(mem_id)     # adds None to seen_ids
```

After the first memory with `id=None` is processed, `None` is added to `seen_ids`. Every subsequent memory with `id=None` hits `None in seen_ids` and is **silently discarded**.

### Impact
- **Data loss**: Multiple memories without ids are reduced to just one
- **Silent**: No error, no warning, no log -- memories just disappear
- **Triggerable**: Any malformed backend data or external documents without ids

### Fix
Only deduplicate when the id is not None:
```python
if mem_id is not None and mem_id in seen_ids:
    continue
seen_ids.add(mem_id)
```

Memories without an id cannot be deduplicated, so they should always be included.

### Test
Added `tests/test_search_changed_since_none_id.py` with two regression tests:
1. `test_multiple_none_id_memories_not_dropped` -- verifies all id=None memories are kept
2. `test_duplicate_real_ids_still_deduped` -- verifies real duplicate ids are still deduplicated

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Memories without an ID are now retained and processed correctly instead of being mistakenly treated as duplicates.
  * Duplicate memories with valid IDs continue to be filtered out.

* **Tests**
  * Added regression coverage for ID-less memories and duplicate valid IDs.


windy202508:

Thanks for the thorough review! Both issues have been fixed:

### 1. Preserve missing-ID memories in `_fetch_all_memories` ✅

The root cause was deeper than the initial fix — `_fetch_all_memories` was dropping all memories with `id=None` at line 621 (`if not mid: continue`) before they ever reached `search_changed_since`.

**Fix:** Memories without an id are now collected in a separate `no_id_memories` list and appended to the results after deduplication. This ensures:
- Identified memories are still deduplicated by version (newest wins)
- Missing-ID memories are retained without being silently dropped
- `search_changed_since` receives the full memory set

### 2. Fixed test API usage ✅

Tests have been rewritten to:
- Use the synchronous `search_changed_since(since_date=...)` API (no `await`)
- Mock `_get_search_namespaces` and `_fetch_all_memories` (the actual methods used by the implementation)
- Add `created_at` timestamps so memories qualify as "changed"
- Assert on `result["results"]` instead of `len(result)`

Both tests now properly exercise the fix path.

Refs #770

Xenogents:

PR #1819, opened 2026-08-05
fix: use versioned marker to distinguish serialized tags from user content

windy202508:
## Problem

When a memory record has tags, the serializer appends a trailing `\n\nTags: tag1, tag2` block to the content. The reader strips this block on readback.

However, if the user's actual content ends with `\n\nTags: python, ai` and the record's tags are `["python", "ai"]`, the reader cannot distinguish between serialized metadata and legitimate user content — it silently strips the user's content.

### Serializer (`core.py`)
Add a unique, versioned marker before the tags block:
```python
# Before: text += f"\n\nTags: {', '.join(self.tags)}"
# After:  text += f"\n\n<!--memanto-tags:v1-->\nTags: {', '.join(self.tags)}"
```

### Reader (`memory_read_service.py`)
Check for the marker first:
- If `<!--memanto-tags:v1-->` is present → strip from marker onwards (definitely serialized metadata)
- If marker not found → fall back to old behavior for backward compatibility (check `Tags: ` prefix + verify tags match)

## Test

Added regression test in `tests/test_format_memory_tags_bug.py` covering:
1. User content ending with `Tags: python, ai` is preserved (not stripped)
2. Serialized tags block with marker is correctly stripped
3. Old-format records (without marker) still work

## Backward Compatibility

Old records without the marker are handled by the fallback path, which uses the same tag-matching logic as before. No data loss for existing records.

Refs #770

> Note: This is a rebase of PR #1810 onto current main to resolve merge conflicts.

### DIFF
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -113,7 +113,7 @@ class MemoryRecord(BaseModel):
         # Format text as standardized card for semantic search
         text = f"[{memory_type.upper()}] {self.title}\n\n{self.content}"
         if self.tags:
-            text += f"\n\nTags: {', '.join(self.tags)}"
+            text += f"\n\n<!--memanto-tags:v1-->\nTags: {', '.join(self.tags)}"
 
         # Build document with flat metadata fields (not nested!)
         document = {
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #1822, opened 2026-08-06
fix: daily summary must fail loudly on empty/missing AI answer (Bounty #770 round 6)

liushazyy:
### Bug
`DailyAnalysisService.generate_summary` did:
```python
summary_text = result.get("answer", "Failed to generate summary.")
```
When the backend returns an **empty** or **missing** `answer` (rate limit, model error, malformed response), the placeholder text (or empty string) was written to the day's summary file and the method returned `{"status": "success", ...}` — a silent failure indistinguishable from a real daily summary.

The legacy summarizer (`context_summarization_service.py:112`) already raises `MemoryError("Failed to generate summary - empty response from AI")`; the new service lost that guard.

### Fix
- Empty/missing/whitespace-only answer now raises `MemoryError` before any file I/O
- `except MemoryError: raise` added so the guard isn't re-wrapped
- No summary file is written on failure (no fake "success" artifact on disk)

### Tests (`tests/failing_tests/test_bounty_770_round6.py`)
- `test_empty_answer_must_not_report_success`: `{"answer": ""}` -> MemoryError, no file written
- `test_missing_answer_key_must_not_write_placeholder`: `{"foo": "bar"}` -> MemoryError, no file written
- `test_valid_answer_still_writes_summary`: real answer still produces the file (no regression)

Verified RED (2 failures without fix) -> GREEN (all pass with fix).

## Summary by CodeRabbit

* **Bug Fixes**
  * Tags containing commas are now rejected to preserve reliable storage and retrieval.
  * Invalid recall limits and similarity settings now return clear validation errors.
  * Search result totals now reflect filtered matches accurately.
  * Memory retrieval avoids stalled or excessive pagination.
  * Missing AI summaries now raise an error instead of creating placeholder files.
  * Conflict reports and exports now respect the configured data directory.
  * Tag handling is consistent for both list and comma-separated formats.

* **Tests**
  * Added regression coverage for validation, pagination, search totals, tag handling, summaries, and configured storage paths.

### DIFF
--- a/memanto/app/core.py
+++ b/memanto/app/core.py
@@ -16,7 +16,18 @@ from memanto.app.constants import (
 )
 
 MemoryTag = Annotated[
-    str, StringConstraints(strip_whitespace=True, min_length=1, max_length=64)
+    str,
+    StringConstraints(
+        strip_whitespace=True,
+        min_length=1,
+        max_length=64,
+        # Commas are the wire-format separator: to_moorcheh_document joins
+        # tags with "," and readers split on ",". A tag containing a comma
+        # is silently corrupted on the write->read round-trip (the tag is
+        # split into two, a phantom tag appears) and makes #tag filtering
+        # match the wrong rows. Reject it at the schema boundary.
+        pattern=r"^[^,]+$",
+    ),
 ]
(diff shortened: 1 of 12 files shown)


Xenogents:

PR #1826, opened 2026-08-06
fix: preserve expires_at on memory update when ttl_seconds is absent

windy202508:
### Summary

When updating any field on a memory that has `expires_at` set but no `ttl_seconds`, the expiration is silently removed, making the memory permanent. This is a **silent data loss** bug.

### Root Cause

In `memory_write_service.py`, the TTL handling in `update_memory()`:

```python
if "ttl_seconds" in updates:
    updated_memory.set_ttl(updates["ttl_seconds"])
elif metadata.get("ttl_seconds"):
    updated_memory.ttl_seconds = metadata["ttl_seconds"]
    raw_expires_at = metadata.get("expires_at")
    if raw_expires_at:
        updated_memory.expires_at = ...
```

When the existing memory has `expires_at` but **no** `ttl_seconds`:
1. `"ttl_seconds" not in updates` → skip first branch
2. `metadata.get("ttl_seconds")` → `None` (falsy) → skip elif branch
3. `updated_memory.expires_at` stays at default (`None`)
4. **The memory's expiration is silently removed**

This can happen when `expires_at` is set via direct metadata write, or when a memory was imported with an explicit expiration but no TTL duration.

### Fix

Add an `else` branch that preserves `expires_at` from the existing metadata when `ttl_seconds` is not present:

```python
else:
    # Preserve expires_at when ttl_seconds is absent (bounty #770).
    raw_expires_at = metadata.get("expires_at")
    if raw_expires_at:
        if isinstance(raw_expires_at, str):
            try:
                updated_memory.expires_at = datetime.fromisoformat(
                    raw_expires_at.replace("Z", "+00:00")
                )
            except (ValueError, AttributeError):
                pass
        else:
            updated_memory.expires_at = raw_expires_at
```

### Impact

- **Severity**: High — silent data loss
- **Type**: Logic error (missing code path)
- **Affected operations**: Any `PATCH /{agent_id}/memories/{memory_id}` on a memory with `expires_at` but no `ttl_seconds`
- **User impact**: Memories that should expire become permanent without any warning or error

### Testing

Added regression test `tests/test_update_memory_preserves_expires_at.py` covering:
1. Update with `expires_at` but no `ttl_seconds` → `expires_at` preserved
2. Update with both `expires_at` and `ttl_seconds` → both preserved
3. Update with new `ttl_seconds` → `expires_at` recalculated correctly

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory updates now preserve existing expiration settings when no new time-to-live value is provided.
  * Existing expiration timestamps remain intact, including valid timestamp values supplied as strings.
  * Providing a new time-to-live value correctly recalculates the expiration time.

* **Tests**
  * Added coverage for preserving and recalculating expiration settings during memory updates.

### DIFF
--- a/memanto/app/services/memory_write_service.py
+++ b/memanto/app/services/memory_write_service.py
@@ -388,6 +388,22 @@ class MemoryWriteService:
                             pass  # Keep the default if the stored timestamp is invalid
                     else:
                         updated_memory.expires_at = raw_expires_at
+            else:
+                # Preserve expires_at when ttl_seconds is absent (bounty #770).
+                # A memory can have expires_at set without ttl_seconds (e.g. via
+                # direct metadata write). Without this branch, any field update
+                # silently drops the expiration, making the memory permanent.
+                raw_expires_at = metadata.get("expires_at")
+                if raw_expires_at:
+                    if isinstance(raw_expires_at, str):
+                        try:
+                            updated_memory.expires_at = datetime.fromisoformat(
+                                raw_expires_at.replace("Z", "+00:00")
+                            )
+                        except (ValueError, AttributeError):
+                            pass
+                    else:
+                        updated_memory.expires_at = raw_expires_at
 
             # Step 3: Upload new version (overwrites existing document with same ID)
             from typing import Any, cast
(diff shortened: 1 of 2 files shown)


Xenogents:

PR #1827, opened 2026-08-07
fix: memory-integrity fixes — extraction truncation, summary archiving, JWT payload validation, session duration, export timestamps (Bounty #770 round 7)

liushazyy:
### 1. _conversation_text truncated the NEWEST messages, or returned empty
conversation_memory_extraction_service.py built the extraction query from the oldest messages first and broke out once the budget was exceeded — dropping exactly the turns this request is about. A single message over the budget produced an empty query (caller raises 400) on input that should be accepted. Fix: walk from the newest message backwards; a single oversized message is kept rather than yielding an empty query.

### 2. Session summaries archived by memory.created_at — imported memories vanish from daily summaries
log_memory_to_session_summary filed entries under the memory's created_at date. Imported/backfilled memories with an old created_at were written to a historical file that daily_analysis_service (which globs today's date) never reads — today's activity permanently missing, and deletion entries landed in a different file than the memory they refer to. Fix: archive under the write time (utc_now()).

### 3. validate_session leaked ValidationError → HTTP 500 on signature-valid but malformed payloads
A JWT signed with the correct key but missing required fields raised pydantic.ValidationError — uncaught, escaping as a 500. Fix: map ValidationError/TypeError/ValueError to InvalidSessionTokenError (4xx).

### 4. Export header used naive local time mixed with UTC memory timestamps
format_memory_md wrote Generated: with datetime.now() (naive local) while every memory timestamp is UTC ISO — two time bases in one exported file, off-by-one-day near midnight UTC+8. Fix: datetime.now(timezone.utc) ISO with Z suffix.

### 5. duration_hours <= 0 minted already-expired sessions while the CLI reported success
_create_session accepted 0/negative hours → token expired at its own start instant; memanto agent activate --hours 0 printed OK but every subsequent call failed. Fix: reject non-positive durations at the service layer (ValueError) and add min=1 to the CLI option.

### Verification
- tests/failing_tests/test_bounty_770_round7.py: 6 tests, all red before / green after
- Full suite: 590 passed, 26 skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Conversation context now prioritizes the newest messages while preserving chronological order.
  * Oversized messages no longer produce empty conversation queries.
  * Session creation rejects invalid or non-positive durations.
  * Malformed session tokens now return a consistent invalid-token error.
  * Memory summary archives use the correct UTC write date.
  * Export timestamps now use UTC ISO 8601 formatting.
  * Agent activation requires at least one hour.

* **Tests**
  * Added regression coverage for conversation trimming, session validation, archiving, and UTC timestamps.

### DIFF
--- a/memanto/app/services/conversation_memory_extraction_service.py
+++ b/memanto/app/services/conversation_memory_extraction_service.py
@@ -98,12 +98,25 @@ class ConversationMemoryExtractionService:
     def _conversation_text(self, messages: list[dict[str, str]]) -> str:
         lines: list[str] = []
         total = 0
-        for message in messages:
+        # Messages arrive in chronological order; the newest turns are the
+        # ones this extraction request is actually about. Walk from the END
+        # so truncation drops the oldest messages first and the query always
+        # ends with the latest context. Without this, an oversized transcript
+        # was cut from the front, leaving a query of stale messages — or, for
+        # a single oversized message, an empty string that made the caller
+        # raise a 400 on input that should be accepted.
+        for message in reversed(messages):
             line = f"{message['role'].strip()}: {message['content'].strip()}"
-            total += len(line)
-            if total > self.MAX_CONTENT_CHARS:
+            if lines and total + len(line) > self.MAX_CONTENT_CHARS:
+                # Budget exhausted before this (older) message: stop.
                 break
             lines.append(line)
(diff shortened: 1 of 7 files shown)


Xenogents:

PR #1828, opened 2026-08-07
fix: langgraph store crash + silent memory drop, ui_router bad-input 400s, CLI migrate friendly errors (Bounty #770 round 8)

liushazyy:
### 1. LangGraph MemantoStore._do_list_namespaces crashed on every call (AttributeError)
SdkClient.list_agents() returns the AgentList envelope dict ({'agents': [...], 'count': N, 'warnings': [...]}), but the store iterated the dict directly — yielding its string keys and calling .get() on a str. Every namespace-listing operation crashed. Fix: unwrap the 'agents' list defensively (accepting both envelope and bare-list shapes). Test: test_do_list_namespaces_handles_list_agents_envelope.

### 2. LangGraph create_remember_node silently dropped oversized memories
Content over the 10k-char memory cap made SdkClient.remember raise ValueError, which the node's except Exception swallowed — logging an error and returning {'messages': []} as if the write had succeeded. Memory lost, caller unaware. Fix: truncate content to the cap (keeping the newest text) before the call so the memory is actually stored. Test: test_remember_node_oversized_content_fails_loudly.

### 3. ui_router migration: existing-but-broken export JSON returned HTTP 500
_migrate_load_or_export checked file existence but let JSONDecodeError escape for a corrupt file — a user-input error surfaced as a server fault. Fix: map JSONDecodeError/OSError/ValueError to HTTP 400. Test: test_migrate_load_or_export_broken_json_is_400.

### 4. ui_router endpoints crashed with AttributeError on non-string body values
body={'api_key': 12345} (or a non-string provider) called .strip() on an int -> AttributeError -> 500. The recall branch was guarded (round 3) but update_api_key / migrate_dry_run / migrate_import / on-prem answer were not. Fix: _coerce_str() helper rejects non-scalar types with 400 and stringifies scalars. Test: test_update_api_key_rejects_non_string.

### 5. CLI migrate --file raised raw FileNotFoundError/JSONDecodeError tracebacks
Unlike migrate_okf and ui_router, the CLI _load_or_export had no exists/parse guard: a missing or corrupt --file dumped a full traceback. Fix: wrap in try/except and raise ValueError (the CLI converts to a friendly _error). Test: test_migrate_load_or_export_missing_file.

### Verification
- tests/failing_tests/test_bounty_770_round8.py: 5 tests, all red before / green after
- Full suite: 589 passed, 26 skipped

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented oversized memories from being rejected during storage.
  * Improved handling of namespace responses from integrations.
  * Added validation for non-text UI inputs and invalid migration data.
  * Replaced unclear migration file errors with user-friendly messages.
  * Invalid requests now return clear HTTP 400 responses.

* **Tests**
  * Added regression coverage for memory limits, namespace handling, input validation, and migration errors.

### DIFF
--- a/integrations/langgraph/langgraph_memanto/nodes.py
+++ b/integrations/langgraph/langgraph_memanto/nodes.py
@@ -246,6 +246,16 @@ def create_remember_node(
             return {"messages": []}
 
         content = "\n\n".join(messages_to_remember)
+        # The memory store caps content at InputLimits.MAX_TEXT_LENGTH (10k
+        # chars) and rejects anything larger with a ValueError. Truncate here
+        # (keeping the newest text, i.e. the end of the joined content) so a
+        # long pasted document or code block is stored instead of silently
+        # dropped — the previous behavior swallowed the ValueError in the
+        # except below, logged an error, and returned {"messages": []} as if
+        # the write had succeeded.
+        max_content = 10_000
+        if len(content) > max_content:
+            content = content[-max_content:]
         title = content if len(content) <= 50 else content[:47] + "..."
 
         agent_client, agent_lock = _cache.get(resolved_agent_id)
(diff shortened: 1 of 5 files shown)


Xenogents:

PR #1834, opened 2026-08-07
fix: use session.session_id instead of hardcoded 'unknown' in CLI client remember/batch_remember

windy202508:
### Problem

In both `SdkClient` and `DirectClient`, the `remember()` and `batch_remember()` methods hardcode `session_id = "unknown"` when logging memories to session summary markdown files, instead of using the actual session ID from the validated session object.

### Impact

1. **Session summary files are misnamed**: `log_memory_to_session_summary` uses `session_id` to construct the filename (`{agent_id}_{date}_{session_id}_summary.md`). With `"unknown"`, all session summaries get the same broken filename regardless of which session they belong to.

2. **Session context is lost**: Memories cannot be traced back to their originating session, breaking the session-to-memory audit trail.

3. **Data corruption on multi-session workloads**: If an agent has multiple sessions, all memories are logged to the same `"unknown"` summary file, mixing session contexts.

### Root Cause

In `SdkClient.remember()`:
```python
session = self._get_validated_session_for_agent(agent_id)
_ = session  # session is obtained but DISCARDED
# ...
session_id = "unknown"  # should be session.session_id
```

In `SdkClient.batch_remember()`:
```python
self._get_validated_session_for_agent(agent_id)  # result not captured
# ...
session_id = "unknown"  # should be session.session_id
```

Same pattern in both `DirectClient.remember()` and `DirectClient.batch_remember()`.

### Fix

- Preserve the session reference instead of discarding it
- Use `session.session_id` instead of `"unknown"`

### Comparison with correct code

The `delete_memory()` method in both clients correctly uses `session.session_id`:
```python
session = self._get_validated_session_for_agent(agent_id)
# ...
session_id=session.session_id  # correct!
```

And the FastAPI route handlers also correctly use `session.session_id`:
```python
session: Session = Depends(get_current_session)
# ...
session_id=session.session_id  # correct!
```

### Tests

Added `tests/test_session_id_logging.py` with regression tests for all four methods.

Refs #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Session summaries now record the correct session ID after successful single or batch memory writes.
  * Removed incorrect “unknown” session ID entries.

* **Tests**
  * Added coverage for session ID logging across direct and SDK-based memory operations.

### DIFF
--- a/memanto/cli/client/direct_client.py
+++ b/memanto/cli/client/direct_client.py
@@ -639,7 +639,7 @@ class DirectClient:
         """
         # Ensure there is a valid, non-expired session for this agent
         session = self._get_validated_session_for_agent(agent_id)
-        _ = session
+        # session used below for session_id logging
 
         self._validate_memory_input(memory_type, title, content, confidence)
 
@@ -673,7 +673,7 @@ class DirectClient:
 
         # Log to local session Markdown summary only after a durable write.
         if self.session_token and is_successful_write_result(result):
-            session_id = "unknown"
+            session_id = session.session_id
             self._get_session_service().try_log_memory_to_session_summary(
                 agent_id=agent_id,
                 session_id=session_id,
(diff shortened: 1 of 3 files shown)


Xenogents:

PR #1867, opened 2026-08-16
fix: correct confidence pairing in summary visualization service

windy202508:
### Problem

The `_parse_session_files` method in `summary_visualization_service.py` was matching headings and confidences by **index**, using two independent regex scans:

```python
headings = list(self._HEADING_RE.finditer(text))
confidences = list(self._CONFIDENCE_RE.finditer(text))

for i, match in enumerate(headings):
    ...
    if i < len(confidences):
        conf = float(confidences[i].group(1))
```

The heading regex requires a non-empty title (`.+`), but the confidence regex matches any `**Confidence**` line. When a memory has an **empty title**, the heading regex skips it while the confidence regex still matches its confidence line. This causes **all subsequent confidence pairings to be off by one**, producing incorrect confidence statistics in the visualization.

### Example

Session file with 3 memories where the first has an empty title:

```
### [2026-01-01 10:00:00] [FACT] 
- **Confidence**: `0.5`
### [2026-01-01 11:00:00] [FACT] Memory B
- **Confidence**: `0.9`
```

**Before fix**: Only 1 heading matched (Memory B), but 2 confidences matched. Memory B gets paired with confidence 0.5 (wrong, should be 0.9).

**After fix**: Memory B is correctly paired with its nearest following confidence 0.9.

### Fix

Replace index-based pairing with **sequential line-by-line parsing** that matches each heading with its nearest following confidence value (within 10 lines). This ensures correct pairing regardless of whether the heading regex matches or not.

### Tests

Added comprehensive tests covering:
- Normal pairing (headings and confidences in order)
- Empty title (heading skipped, confidence still correct for next entry)
- DELETED entries (have confidence, pair correctly)
- Missing confidence (uses default 0.8)
- Multiple session files

Ref #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session parsing to correctly associate confidence values with their nearest valid headings.
  * Skipped invalid timestamps and malformed confidence values without disrupting other entries.
  * Preserved the default confidence value when no valid confidence is provided.

* **Tests**
  * Added coverage for confidence pairing, empty titles, deleted entries, and missing-confidence scenarios.

### DIFF
--- a/memanto/app/services/summary_visualization_service.py
+++ b/memanto/app/services/summary_visualization_service.py
@@ -123,6 +123,11 @@ class SummaryVisualizationService:
 
         Returns a list of dicts:
             {"timestamp": datetime, "type": str, "title": str, "confidence": float}
+
+        Bug fix (ref #770): Parse headings and confidences sequentially by scanning
+        the text line by line, matching each heading with its nearest following
+        confidence. This avoids index-based pairing which can misalign when
+        the heading regex fails to match (e.g., empty titles).
         """
         pattern = f"{agent_id}_{date}_*_summary.md"
         session_files = list(sessions_dir.glob(pattern))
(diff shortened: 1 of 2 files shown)


Xenogents:
This is a duplicate.

PR #1878, opened 2026-08-19
fix: use actual session_id instead of hardcoded 'unknown' in DirectClient session summary logging

windy202508:
## Bug Description

In `DirectClient.remember()` and `DirectClient.batch_remember()`, the `session_id` was hardcoded to `"unknown"` when logging memories to session summary Markdown files.

### Impact

This caused several problems:

1. **Incorrect file naming**: Session summary files were named `{agent_id}_{date}_unknown_summary.md` instead of using the real session ID (e.g., `sess_abc123`)

2. **Broken daily summary & conflict report**: `DailyAnalysisService.generate_summary()` and `generate_conflict_report()` search for files matching the pattern `{agent_id}_{date}_*_summary.md`. While the glob still matches, the session ID embedded in the filename is wrong, making it impossible to distinguish which session produced which memories.

3. **Cross-session data merge**: When multiple sessions exist for the same agent on the same day, all session summary logs are written to the same `unknown` file instead of separate per-session files, merging unrelated session data.

### Root Cause

The `remember()` method calls `self._get_validated_session_for_agent(agent_id)` which returns a valid `Session` object with the correct `session_id`. However, when logging to the session summary, the code used `session_id = "unknown"` instead of `session.session_id`.

The same issue existed in `batch_remember()`.

### Fix

Use `session.session_id` (the actual session ID from the validated session object) instead of the hardcoded `"unknown"` string.

The `session` object is already available from the `_get_validated_session_for_agent()` call at the top of both methods, so no additional API calls or state changes are needed.

### Testing

Added `tests/test_direct_client_session_id.py` with 4 test cases:
- `test_remember_uses_actual_session_id`: Verifies `remember()` passes the real session_id
- `test_batch_remember_uses_actual_session_id`: Verifies `batch_remember()` passes the real session_id
- `test_remember_does_not_use_unknown_session_id`: Ensures `remember()` does NOT use "unknown"
- `test_batch_remember_does_not_use_unknown_session_id`: Ensures `batch_remember()` does NOT use "unknown"

References #770

## Summary by CodeRabbit

* **Bug Fixes**
  * Session summaries now record the correct session ID when saving individual or batches of memories.
  * Eliminated incorrect `"unknown"` session identifiers in local summary logs.

* **Tests**
  * Added coverage for session ID logging across supported client workflows.
  * Added regression checks for both individual and batch memory operations.

### DIFF
--- a/memanto/cli/client/direct_client.py
+++ b/memanto/cli/client/direct_client.py
@@ -639,7 +639,7 @@ class DirectClient:
         """
         # Ensure there is a valid, non-expired session for this agent
         session = self._get_validated_session_for_agent(agent_id)
-        _ = session
+        # session used below for session_id logging
 
         self._validate_memory_input(memory_type, title, content, confidence)
 
@@ -673,7 +673,7 @@ class DirectClient:
 
         # Log to local session Markdown summary only after a durable write.
         if self.session_token and is_successful_write_result(result):
-            session_id = "unknown"
+            session_id = session.session_id
             self._get_session_service().try_log_memory_to_session_summary(
                 agent_id=agent_id,
                 session_id=session_id,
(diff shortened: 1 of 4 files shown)


Xenogents:
Please stop opening multiple PR's for the same bounty.

Issue #1852, opened 2026-08-12
[BOUNTY $100]🛡️The Memanto Security Challenge: Test Our Defenses


Xenogents:
## Description:

In a previous bounty, we asked the community to stress-test **Memanto** in order to find bugs, logic loops, and performance bottlenecks. We got an unexpectedly large number of PR's and we’ve spent a great deal of time meticulously reviewing them, you’ve all helped us make a huge step forward in improving the security and quality of Memanto!

Now, we are taking it a step further and focusing exclusively on what matters most for AI in production: **Security**.

As memory layers become the backbone of autonomous agents, they also become a prime target for attacks. Whether it's a cross-tenant data leak, a broken access control, or a malicious payload hidden in a retrieved memory, we want to know about it. We’ve fixed a majority of bugs and issues with Memanto and resolved many of its security issues, but as Memanto starts to be utilized in more sensitive workflows, we want to ensure our defenses are as robust as possible and we need the developer community to help us do that.

Your mission is simple: **Help us audit our memory management and backend infrastructure to uncover potential vulnerabilities before they can be exploited.** 

---

## 🎯 Threat Model & Scope
Your target is the `memanto` core package and its interaction with the `moorcheh.ai` serverless backend. We are prioritizing traditional application security flaws, data isolation failures, and API vulnerabilities. 

### 🔴 Security Focus Areas (In Scope)
- **Tenant Isolation & Cross-Account Leaks:** Can User A query, retrieve, or overwrite memories belonging to User B through the core package or backend API?
- **Authentication & Authorization Bypasses:** Finding ways to read or write to the memory layer without a valid API key, or bypassing intended role/access restrictions.
- **API Vulnerabilities:** Standard backend flaws such as Insecure Direct Object References (IDOR), broken object level authorization, or injection flaws in the `moorcheh.ai` backend.
- **Data Exfiltration:** Tricking the system or bypassing controls to extract raw vector data, plaintext logs, or sensitive hidden context from the memory store.
- **AI-Specific Exploits (Indirect Prompt Injections):** Exploiting the retrieval process to inject dormant malicious payloads into memory that later hijack the agent's core instructions when recalled (i.e., treating memory as a trojan horse).

### 🚫 Rules of Engagement (Out of Scope)
- Volumetric DDoS attacks against `moorcheh.ai` infrastructure.
- General LLM hallucinations or standard logic bugs unrelated to security.
- Attacks that require compromising the underlying operating system or Docker host to succeed.

---

[...]

PR #1891, opened 2026-08-21
security: fail closed on tenant-scoped memory reads (bounty #1852)

laurentketterle-hub:
## Summary

Fixes a **cross-tenant data isolation failure** in the `memanto` core read service, submitted for the Memanto Security Challenge (#1852).

`MemoryReadService` exposes two entry points whose default behaviour is to *drop* tenant scope and fan out across **every** namespace on the server's Moorcheh account when an `agent_id` is missing or empty:

- `_get_search_namespaces(agent_id=None)` -> `list_namespaces()` -> searches **all** `memanto_agent_*` namespaces (used by `search_memories` / temporal recall / recent recall).
- `generate_answer(query, agent_id=None)` -> `namespaces[0]` -> answers from the **first** namespace in the account list.

Both `agent_id` parameters default to `None`, so a single omitted argument (the default value) silently escalates an agent-scoped read into a read of every other tenant's memories — the exact "cross-tenant data leak" class this challenge asks us to find.

## Threat model fit

- **Tenant Isolation & Cross-Account Leaks** (in scope): a caller that omits `agent_id` can retrieve memories belonging to other agents/tenants sharing the same server account.

## Reproduction (PoC)

```python
from unittest.mock import MagicMock
from memanto.app.services.memory_read_service import MemoryReadService

client = MagicMock()
client.namespaces.list.return_value = {
    "namespaces": [
        {"namespace_name": "memanto_agent_alice"},
        {"namespace_name": "memanto_agent_bob"},
    ]
}
client.similarity_search.query.side_effect = (
    lambda query=None, namespaces=None, **kw:
        {"results": [{"id": f"mem_of_{ns}", "text": f"secret of {ns}"}
                      for ns in (namespaces or [])]}
)
client.answer.generate.side_effect = (
    lambda namespace=None, **kw: {"answer": f"from {namespace}", "sources": []}
)

svc = MemoryReadService(client)

# Before this patch: returns BOTH tenants' memories
print(svc.search_memories(query="anything", agent_id=None, limit=10))
# Before this patch: answers from "memanto_agent_alice" (first tenant)
print(svc.generate_answer(query="anything", agent_id=None))
```

## Fix

Fail closed instead of fail open. Both paths now require an explicit, non-empty `agent_id` and raise `MemoryError` otherwise, so a read can never silently expand beyond the caller's own namespace.

## Changes

- `memanto/app/services/memory_read_service.py`: `_get_search_namespaces` and `generate_answer` now require an `agent_id` (no `list_namespaces()` / `namespaces[0]` fallback).
- `tests/test_memory_read_tenant_isolation.py`: regression tests covering the all-namespaces fan-out and the first-namespace answer fallback.

## Verification

- `pytest tests/test_memory_read_tenant_isolation.py` -> 4 passed
- `pytest tests/` (excluding network e2e/api) -> all green, no regressions

Closes the data-isolation gap described in #1852.

## Summary by CodeRabbit

* **Bug Fixes**
  * Memory searches and answer generation now require a non-empty agent identifier.
  * Results are restricted to the requesting agent’s namespace, preventing cross-tenant access and fallback.
  * Minimum-confidence filtering now safely retains records with malformed or non-finite confidence values.

* **Tests**
  * Added coverage for missing identifiers, namespace isolation, cross-tenant search prevention, and confidence filtering behavior.

### DIFF
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -2,6 +2,7 @@
 Memory Read Service
 """
 
+import math
 import re
 from concurrent.futures import ThreadPoolExecutor
 from datetime import datetime, timezone
@@ -810,6 +811,11 @@ class MemoryReadService:
                 # Fail open: include memories with unknown confidence rather
                 # than silently dropping them. This preserves imported memories
                 # that may not carry a confidence score.
+                filtered.append(result)
+                continue
+            if not math.isfinite(confidence):
+                # Non-finite values (NaN/inf) are malformed: fail open, same as
+                # unparseable confidence above, rather than silently dropping.
                 filtered.append(result)
                 continue
             if confidence >= min_confidence:
--- a/memanto/app/services/memory_read_service.py
+++ b/memanto/app/services/memory_read_service.py
@@ -847,15 +847,15 @@ class MemoryReadService:
     ) -> dict[str, Any]:
         """Generate AI answer from memories"""
         try:
-            # Determine namespace for answer generation
-            if agent_id:
-                namespace = agent_namespace(agent_id)
-            else:
-                # Use first available namespace
-                namespaces = self.namespace_service.list_namespaces()
-                if not namespaces:
-                    raise MemoryOperationError("No namespaces found")
-                namespace = namespaces[0]
+            # Tenant isolation: answer generation must be scoped to one agent's
+            # namespace. The previous "first available namespace" fallback let a
+            # caller without an ``agent_id`` read memories from whichever tenant
+            # happened to sort first in the account's namespace list.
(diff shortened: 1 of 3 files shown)


laurentketterle-hub:
Addressed the CodeRabbit feedback: non-finite confidence values (NaN/inf) now follow the documented fail-open path instead of being silently dropped, with a regression test added. Ready for review when you get a chance — thanks!

laurentketterle-hub:
Gentle follow-up — the CodeRabbit feedback was addressed on Aug 28 (non-finite confidence values now follow the documented fail-open path, plus a regression test). Could you review when you have a moment? Thanks!

PR #1997, opened 2026-09-16
fix(parsing): let stated intent outrank a topical error keyword

### LINKED ISSUE #1890: MemoryParsingService classifies error-adjacent preferences and instructions as `error`
*Found while building a migration showcase for #1609. Imported memories are
auto-classified in bulk, which is how the pattern became visible.*

## Summary

`MemoryParsingService` classifies a preference or instruction as `error`
whenever it happens to mention errors, failures, or bugs. The rules match the
*topic* of errors rather than the *reporting* of one, so "how I want errors
presented" lands in the same bucket as "an error occurred".

## Reproduction

Rule-based path only, so no API key and no network:

```python
from memanto.app.core import MemoryRecord
from memanto.app.services.memory_parsing_service import MemoryParsingService

svc = MemoryParsingService()
for text in [
    "Show the failing case first, then the fix.",
    "Prefer answers that explain what went wrong before the solution.",
    "I want error messages explained in plain English.",
    "When reviewing my code, always point out the bug before suggesting a fix.",
    "Always run the failing test before changing anything.",
]:
    got = svc.parse_memory(MemoryRecord(
        title=text[:100], content=text,
        agent_id="a", actor_id="a", source="user")).type
    print(f"{got:<12} {text}")
```

Output on `main` (c84429c):

```
error        Show the failing case first, then the fix.
error        Prefer answers that explain what went wrong before the solution.
error        I want error messages explained in plain English.
instruction  When reviewing my code, always point out the bug before suggesting a fix.
instruction  Always run the failing test before changing anything.
```

The first three are a user preference or standing instruction, not an incident.

## Root cause

Classification takes the single strongest rule score. The `error` ruleset scores
**5** for a bare word match:

```python
r"\b(?:errors?|fail(?:s|ed|ing|ure)?|bugs?|exceptions?|tracebacks?|crash(?:e[ds]|ing)?|outages?|incidents?)\b"
```

For `"i want error messages explained in plain english."` that is the *only*
rule that fires at all, so `error` wins uncontested. There is no competing
signal for a leading "I want ..." or "Prefer ...".

Note the last two cases classify correctly, so this is not "the error rules are
too broad" in general. It is specifically that a topical mention wins when no
intent phrase competes with it.

## Impact

Memory type drives retrieval filtering (`#memory_type:preference`). A preference
filed under `error` will not surface when an agent asks for preferences, and it
pollutes error queries with things that are not errors.

This compounds during migration. Imported memories are auto-classified in bulk,
and assistant memory is full of sentences about how the user wants problems
explained, so the misfile rate is meaningfully higher on migrated data than on
hand-written memories.

## Possible directions

Deliberately not prescribing a fix, but the options seem to be: require an
error rule to co-occur with a reporting signal (past tense, a stack trace, an
identifier) before scoring 5; lower the bare-keyword score below
`MIN_RULE_SCORE` on its own; or add intent-phrase rules for "I want", "Prefer",
"I like" so preference has something to compete with.

Happy to open a PR if a direction is preferred.

tangjiadev:
Fixes #1890.

The `error` ruleset scores 5 on a bare mention of the word, and nothing competed for it, so a stated preference or a presentation directive got filed as an error. A request like "I want error messages explained in plain English" came out as `error` instead of `preference`.

What changed:

- A `preference` rule for desire framing about how results are reported, scored to tie with that keyword rule.
- An `instruction` rule for a sentence that opens on a presentation directive, scored the same way.
- The tie is resolved by `TYPE_PRIORITY`, which is already how "always" and "never" beat a topical keyword.

Kept narrow on purpose: a "to <verb>" continuation still goes to `goal`, and a desire about the work itself keeps its weaker score, so a hard requirement in the same sentence still wins.

Verified locally:

- 5 new tests in `tests/test_memory_parsing.py`; all five fail on `main` and pass with the fix. 22 tests in the file pass.
- `ruff check` and `ruff format --check` clean on both files.

## Summary by CodeRabbit

- **New Features**
  - Improved recognition of preferences about reporting formats, presentation, summaries, wording, and related outputs.
  - Improved recognition of presentation-focused instructions, including requests to show, explain, format, highlight, list, or summarize information.
  - Better distinguishes goals phrased with “to + verb” from preferences and instructions.

- **Bug Fixes**
  - Prevented work-choice preferences from being misclassified as mandatory requirements.


tangjiadev:
Fixed in ada386f. The alternate branch needed the same `to <verb>` guard as the "I want" branch — "Prefer to ship the CLI by Friday" scored 5 as a preference instead of tying with the goal ruleset at 4, so it went from goal on main to preference here. Both branches carry the guard now, and the goal test covers the three sentence-opening shapes that regressed.

Issue #1948, opened 2026-09-08
Memanto connect skill hardening across different platforms


Xenogents:
## Overview

The memanto connect operation works by injecting markdown (.md) instruction files and skill templates into a project's workspace. Different AI coding agents (VS Code Copilot, Cursor, Windsurf, Claude Code, Roo, etc.) read these markdown files as part of their system prompts. 

By injecting these prompts, we give the agents instructions on how and when to use Memanto's CLI to manage persistent memory. However, injecting the prompt is only the first step. We face two major challenges in making this actually work well across different agents.

---

## Problem 1: Making Memory Operations Actually Get Called

Because the injected rules are just markdown text in a prompt, the LLM has to actively pay attention to them on every single turn. 

*   **The Issue:** Models have a strong tendency to bypass instructional prompts, especially during casual conversation or standard coding tasks. They simply ignore the directives that tell them to evaluate the context and use the memory tools.
*   **The Goal:** We need to craft these prompt instructions in a way that forces the LLM to reliably stop, evaluate the criteria we've given it, and actually invoke the Memanto CLI tools when relevant.

---

## Problem 2: Making It a Good User Experience

To solve Problem 1, we often have to force the LLM to write out its evaluation process step-by-step before it acts. However, if the LLM prints this internal reasoning directly into the chat, it creates a terrible, cluttered user experience.

*   **The Issue:** We need the agent to perform its memory evaluation on every turn, but the user shouldn't have to see a wall of text every time they ask a simple question.
*   **The Current Workflow:** Right now, our template tries to hide this evaluation by asking the LLM to wrap its text in <thinking> blocks, or by forcing it to make a dummy tool call (like echo) and hiding the text inside a tool parameter (like explanation). 
*   **The Goal:** While the current workflow uses <thinking> tags and dummy tool calls, these aren't the only ways to solve this. We need a hardened approach to ensure that whatever method we use, whether it's XML tags, tool parameters, or something else entirely, the internal memory processing remains completely hidden and seamless across all the different agent UIs. If needed, this may involve crafting unique solutions across each different platform.

---

## Current Architecture

Currently, our prompt strategy in memanto/cli/connect/templates.py attempts to address these problems by dividing agent environments into two execution protocols:

1. **Native CLI & Integrated Agent Environments** *(Claude Code, Cursor, Windsurf, Roo, Cline, Continue, Gemini CLI, OpenCode, Codex, Aider, Pi Agent)*:
[...]

PR #1998, opened 2026-09-16
refactor(connect): declare the memory protocol per agent

tangjiadev:
Part of #1948.

Today the injected instruction carries both evaluation-hiding protocols, wrapped in a prose list naming platforms Memanto does not support. Every model has to work out which branch applies to it, and the prompt ends up longer than it needs to be.

What changed:

- `AgentDef` gains a `memory_protocol` field with no default, so a platform has to declare how its UI hides the evaluation. Thirteen agents declare `thinking-block`; `github-copilot` declares `dummy-tool-call` — the same split the templates used to describe in prose.
- The template renders only the agent’s own clause, read from the registry, and the stale platform names are gone.
- `TEMPLATE_VERSION` moves to 1.1.0 so `memanto connect` offers the update.

Verified locally:

- 7 new tests in `tests/test_connect_protocol.py`. Six of them fail on `main` (stashed the fix to check). Registering an agent without declaring a protocol now raises instead of silently inheriting one.
- `ruff check` and `ruff format --check` clean.

## Summary by CodeRabbit

- **Improvements**
  - Connect instructions are now tailored to each supported agent’s interface.
  - Memory evaluation details are presented using the appropriate collapsed reasoning or unobtrusive terminal interaction for the selected agent.
  - Unsupported platform-specific instructions and legacy headings are no longer included.
  - Unrecognized agents use the native reasoning format as a fallback.
  - Connect instruction templates are updated to version 1.1.0.


Xenogents:
Hi @tangjiadev!

Thanks for the PR, I would love to go with this kind of approach to have true separation between instructions between different IDE's and coding agents. My major concern and why it's not implemented that way currently is because I've noticed that IDE's typically look at an array of different md files and a lot of them overlap. In my Vscode copilot setup my agent will look at both my instructions.md file and CLAUDE.md file at once and if they were different then they would receive differing instructions.

Another approach I've thought of is additionally injecting data into ignore files but I figured that may cause more harm if developers deliberately have it set up so multiple md files are read. At the moment, because we only have 2 different approaches so far I feel the current implementation is best but that may be open to change. You're free to comment any of your thoughts on this.

PR #1999, opened 2026-09-16
feat(recall): search several agents in one query

### LINKED ISSUE #542: Add cross-agent / multi-namespace recall
### Problem
Every recall is scoped to a single active agent. Users who run multiple agents (e.g. one per project, or per customer) cannot search across them in one query. 

### Proposed solution
1. **Route** - `POST /api/v2/recall/multi` accepting a list of `agent_ids` plus the usual recall parameters (query, limit, type, min_confidence). Reads through `MemoryReadService` (`memanto/app/services/memory_read_service.py`), Moorcheh support multiple namespace search by passing list of agent namespace.
2. **Auth** - decide and document the auth model: either introduce a multi-agent token or some other approach. Call this out explicitly in the PR.
3. **Client + CLI** - `memanto recall "<query>" --agents agent-a,agent-b,agent-c`.

### Acceptance criteria
- [ ] A single query returns merged, score-sorted results across the listed agents.
- [ ] Each result indicates which agent it came from.
- [ ] Requesting an agent the caller isn't authorized for is rejected cleanly.
- [ ] Existing single-agent `recall` behavior is unchanged.

tangjiadev:
Closes #542.

Recall is scoped to one agent, so anyone running an agent per project, or per customer, cannot search across them in one query.

What changed:

- `MemoryReadService.search_memories_multi` searches each requested agent through the existing `search_memories` path — so filters, ranking and lifecycle handling are unchanged — and merges the hits into one score-ranked page. The per-agent searches are independent, so they are dispatched in parallel.
- `POST /api/v2/recall/multi` exposes it. The auth model, called out as the issue asks: this is the one recall route that is not session-scoped, because a session belongs to a single agent. It therefore takes the management credential (loopback caller, or the server key as `Authorization: Bearer` / `X-Api-Key`), the same trust level as agent create/delete. A per-agent session token cannot be used here, so an agent credential can never read a sibling agent’s memories.
- `memanto recall "<query>" --agents agent-a,agent-b` on the CLI, and both clients learned `recall_multi`.
- Every memory in the response carries `agent_id`. `limit` applies to the merged ranking rather than to each agent, and the same id appearing under two agents stays two memories.

Verified locally:

- 11 new tests in `tests/test_multi_agent_recall.py`: per-agent labelling, cross-agent ranking, page size, duplicate agents, blank ids, a shared id under two agents, parallel dispatch, and request validation. They fail on `main` by construction, since the method and the model do not exist there.
- 65 recall-related tests across `tests/test_api.py`, `tests/test_cli.py` and the read-service suites pass. `ruff check` and `ruff format --check` clean.
- `app.openapi()` lists `POST /api/v2/recall/multi` alongside the existing recall routes.

## Summary by CodeRabbit

- **New Features**
  - Added multi-agent recall, allowing one query to search across multiple agents.
  - Results are merged and ranked together, identify their source agent, and respect a combined result limit.
  - Added CLI support through `--agents`, with duplicate agent IDs removed and accurate result labeling.
  - Multi-agent recall does not require an active agent session.

- **Documentation**
  - Updated CLI and session architecture documentation with multi-agent usage details and examples.


tangjiadev:
Fixed in bd7cc04. The CLI collapses ids the same way the clients already do, so `--agents agent-a,agent-a` counts as one agent in the header. Also made the single-id case read as "1 agent" rather than "1 agents", and added the CLI tests for --agents that were missing.

